PR #107, opened 2025-01-28
Create pl.md

nkkfs:

## Description

Adding Polish translate for Pangolin.
Probably a lot of strings I failed to “call” to translate such as information on successful action and the like.

After the introduction of i18n, I will do my best to translate Pangolin fully.


miloschwartz:
Thanks again for getting this started. I will let you know when we have i18n read to go!

PR #118, opened 2025-01-31
Small Bugfixes

oschwartz10612:

## Description
Small hotfix bug fixes


server/setup/scripts/1.0.0-beta10.ts
oschwartz10612, line 1: @miloschwartz here is the migration to set secure_cookies to true

PR #130, opened 2025-02-01
Update pl.md

nkkfs:
Add Authentication Site strings

### DIFF
--- a/internationalization/pl.md
+++ b/internationalization/pl.md
@@ -1,3 +1,23 @@
+## Authentication Site
+
+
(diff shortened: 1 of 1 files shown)


oschwartz10612:
Thanks again for doing this. I am not sure when we will get to the actual internationalization of the application but I hope we can put this to use soon!

nkkfs:
Sure, the most important thing is to make the application in the principle of operation is working,  internationalization is just _Quality of Life_ change, will be when you think it is appropriate and necessary!
Take your time and huge THANKS for doing that.

PR #286, opened 2025-03-05
Crowdsec-Dashboard-Notification

hhftechnology:


miloschwartz:
Hey, thanks for taking the time to work on integrating this into the UI. 

We’re currently trying to avoid integrating specific third-party add-ons directly into the Pangolin UI. Our goal is to maintain Pangolin as a flexible core product, with add-ons like CrowdSec or GeoBlock existing as complementary but independent components around it.

PR #419, opened 2025-03-28
fix(bandwidth): Record correct megaBytesIn/megaBytesOut

hareland:
## Description
Bandwidth is switching side every time new bandwidth is received

Screenshot 2025-03-28 at 00 27 53
Screenshot 2025-03-28 at 00 27 41

This fixes this issue.

### DIFF
--- a/server/routers/gerbil/receiveBandwidth.ts
+++ b/server/routers/gerbil/receiveBandwidth.ts
@@ -59,8 +59,8 @@ export const receiveBandwidth = async (
                 await trx
                     .update(sites)
                     .set({
-                        megabytesOut: (site.megabytesIn || 0) + bytesIn,
-                        megabytesIn: (site.megabytesOut || 0) + bytesOut,
+                        megabytesIn: (site.megabytesIn || 0) + bytesIn,
+                        megabytesOut: (site.megabytesOut || 0) + bytesOut,
                         lastBandwidthUpdate: new Date().toISOString(),
                         online
                     })
--- a/server/routers/gerbil/receiveBandwidth.ts
+++ b/server/routers/gerbil/receiveBandwidth.ts
@@ -59,8 +59,8 @@ export const receiveBandwidth = async (
                 await trx
                     .update(sites)
                     .set({
-                        megabytesIn: (site.megabytesIn || 0) + bytesIn,
-                        megabytesOut: (site.megabytesOut || 0) + bytesOut,
+                        megabytesOut: (site.megabytesOut || 0) + bytesIn,
+                        megabytesIn: (site.megabytesIn || 0) + bytesOut,
                         lastBandwidthUpdate: new Date().toISOString(),
                         online
                     })


oschwartz10612:
I actually swapped these on purpose in a16e4e2e74d7186572694f5f6a18300ead3d1d53 I think from community feedback.

I think it was because its kind of reversed to what you expect because its reading the in/out of the peer on the Wireguard side in the _cloud_. So _in_ is actually from the site and _out_ is actually to the site. Does that make sense?

I could be convinced to change it back though.

notmeta:
> I actually swapped these on purpose in a16e4e2e74d7186572694f5f6a18300ead3d1d53 I think from community feedback.
> 
> 
> 
> I think it was because its kind of reversed to what you expect because its reading the in/out of the peer on the Wireguard side in the _cloud_. So _in_ is actually from the site and _out_ is actually to the site. Does that make sense?
> 
> 
> 
> I could be convinced to change it back though.

Maybe I'm misunderstanding but I think this is unrelated. I have experienced the same issue where when I refresh the page incoming and outgoing seem to swap places

hareland:
> I actually swapped these on purpose in [a16e4e2](https://github.com/fosrl/pangolin/commit/a16e4e2e74d7186572694f5f6a18300ead3d1d53) I think from community feedback.
> 
> I think it was because its kind of reversed to what you expect because its reading the in/out of the peer on the Wireguard side in the _cloud_. So _in_ is actually from the site and _out_ is actually to the site. Does that make sense?
> 
> I could be convinced to change it back though.

You are probably correct with that, but the issue was that the stored value vs the received value was flipped then added to

Old code:

```
megabytesOut: (site.megabytesIn || 0) + bytesIn,
megabytesIn: (site.megabytesOut || 0) + bytesOut,
```

This creates, well, that next time we call this, we use the value from the other side and adds it, hence leading to the "flip" effect.

I have updated the PR to fit the idea that in=out, the calculation before adding bytesIn/bytesOut is fixed to fit that idea

huzky-v:
Just a +1 confirm this is happening to me and fixed for me, this will also fix the incorrect traffic metrics issue (due to the db in/out flip to db every 10s, which is newt report back the traffic) along with the bandwidth display flip issue stated above. 

I use this change and the result is correct.
```
megabytesOut: (site.megabytesOut || 0) + bytesIn,
megabytesIn: (site.megabytesIn || 0) + bytesOut,
```

But it is tricky to "fix" the existing traffic metrics as the value is not correct in the first hand. (May be resetting to zero in db.sqlite may help?)
```
update sites set bytesIn=0, bytesOut=0;
```

oschwartz10612:
> But it is tricky to "fix" the existing traffic metrics as the value is not correct in the first hand. (May be resetting to zero in db.sqlite may help?)

Yeah maybe we could do this in the migration.

oschwartz10612:
I think I understand now. Thanks @hareland for the contribution!

oschwartz10612:
Oh @hareland can you please update your original PR message or add one below with the CLA please? 

```
```

hareland:
> Oh @hareland can you please update your original PR message or add one below with the CLA please?
> 
> ```
> By creating this pull request, I grant the project maintainers an unlimited,
> perpetual license to use, modify, and redistribute these contributions under any terms they
> choose, including both the AGPLv3 and the Fossorial Commercial license terms. I
> represent that I have the right to grant this license for all contributed content.
> ```


By creating this pull request, I (@hareland) grant the project maintainers an unlimited,

oschwartz10612:
Thank you!

hareland:
Hi again @oschwartz10612 , when is this expected to get released through official channel, or is there additional work i can help with to get this released? Thanks in advance

oschwartz10612:
I think this weekend!

PR #469, opened 2025-04-06
Fix crowdsec config

Lokowitz:

## Description
- removed acquis.yaml because it is outdated [see docs](https://github.com/crowdsecurity/crowdsec/tree/master/docker#acquisition-one-file-per-datasource---recommended)
- added acquis.d with acquise files for traefik and  appsec
- removed crowdsec_log folder because it is not needed
- removed bind mounts because they are wrong and not needed for the basic appsec/traefik bouncers. If someone wants to protect the host system with iptable bouncer and syslog/journalctl logs, there is a PR with a detailed documentation.

This was mentioned in https://github.com/orgs/fosrl/discussions/435.

## Open points
I think there is a command needed to copy the `trafik.yaml` and `appsec.yaml` files to the `/config/crowdsec/acquis.d` directory. But i am so sorry, i don't know how.

Hope this helps a bit.


oschwartz10612:
Thanks so much! I tested and it seemed to work!

PR #477, opened 2025-04-07
Add supporter message feature

grokdesigns:

## Description
Add supporter message from request https://github.com/orgs/fosrl/discussions/460

Screen Recording 2025-04-07 at 14 48 02
*Updated

## How to test?
Rebuild from source and update Docker image.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,6 +33,7 @@
                 "@tanstack/react-table": "8.20.6",
                 "axios": "1.7.9",
                 "better-sqlite3": "11.7.0",
+                "canvas-confetti": "^1.9.3",
                 "class-variance-authority": "0.7.1",
                 "clsx": "2.1.1",
                 "cmdk": "1.0.4",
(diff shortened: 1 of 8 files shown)


oschwartz10612:
Thanks @grokdesigns we think this is really cool! We want to put this somewhere in the project but maybe not where you have it at the bottom of the page. The reason is we were trying to set up the supporter key in a way so that when you purchase a key you kind of remove the watermark and it looks stock so users are not bothered by it.

So I think we just should brainstorm where to put this. Where do you think is best that still shows that you have purchased a key but is out of the way? Maybe in the /admin portal or maybe in the drop down for the user in the top left only if they are in the admin group? Or maybe it can be turned on and off in the config? What do you think?

miloschwartz:
@grokdesigns To add to what @oschwartz10612 said, maybe we could also show the confetti when the user enters a valid supporter key? Really liked that interaction!

grokdesigns:
pangolin_demo

@oschwartz10612 @miloschwartz Thanks for the feedback! Totally understand wanting to keep it more subtle. As you can see in the GIF, the only indication they're a supporter when the user is in the interface is the added orange star next to the app name in the footer. On hover it has a popover thanking them and will still shoot some confetti when clicked. I also added a confetti/fireworks animation when a valid license key is entered. I changed all confetti colors to yellow/orange to match the theme a bit more.

Thoughts?

oschwartz10612:
I think this looks awesome - exactly what we were after! Thank you so much for adding this!

PR #496, opened 2025-04-09
Add invitation management

grokdesigns:

## Description
Screen Recording 2025-04-08 at 18 16 12

Was looking for a way to manage/revoke/etc. invitations and didn't see anything, so I hope that functionality isn't just somewhere I didn't notice! Tested with multiple orgs/invites. Confirmed revoked invitations give user error message when using invite link. Considered moving "Create User" button to invitations page, but probably too much of a change to base functionality. Could duplicate "Create User" on both pages if desired.

## How to test?
Build from source and update Docker image.

### DIFF
--- a/server/auth/actions.ts
+++ b/server/auth/actions.ts
@@ -32,6 +32,8 @@ export enum ActionsEnum {
     listRoles = "listRoles",
     updateRole = "updateRole",
     inviteUser = "inviteUser",
+    listInvitations = "listInvitations",
+    removeInvitation = "removeInvitation",
     removeUser = "removeUser",
     listUsers = "listUsers",
     listSiteRoles = "listSiteRoles",
@@ -63,7 +65,7 @@ export enum ActionsEnum {
     listResourceRules = "listResourceRules",
     updateResourceRule = "updateResourceRule",
     listOrgDomains = "listOrgDomains",
-    createNewt = "createNewt",
+    createNewt = "createNewt"
 }
 
 export async function checkUserActionPermission(
--- a/server/openApi.ts
+++ b/server/openApi.ts
@@ -8,6 +8,7 @@ export enum OpenAPITags {
     Resource = "Resource",
     Role = "Role",
     User = "User",
+    Invitation = "Invitation",
     Target = "Target",
     Rule = "Rule",
     AccessToken = "Access Token"
--- a/server/routers/external.ts
+++ b/server/routers/external.ts
@@ -157,13 +157,6 @@ authenticated.delete(
     user.removeInvitation
 );
 
-authenticated.get(
-    "/org/:orgId/users",
-    verifyOrgAccess,
-    verifyUserHasAction(ActionsEnum.listUsers),
-    user.listUsers
-);
-
 authenticated.post(
     "/org/:orgId/create-invite",
     verifyOrgAccess,
--- a/server/routers/external.ts
+++ b/server/routers/external.ts
@@ -143,6 +143,27 @@ authenticated.get(
     domain.listDomains
 );
 
+authenticated.get(
+    "/org/:orgId/invitations",
+    verifyOrgAccess,
+    verifyUserHasAction(ActionsEnum.listInvitations),
+    user.listInvitations
+);
+
+authenticated.delete(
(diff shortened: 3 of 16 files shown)


grokdesigns:
image
image

Latest commit adds child menu item support to the SidebarNav function, moved Invitations under Users and Invitation menu item will only appear if there are invites in the userInvites table.

Would also like to add ability to regenerate an invite in case link is lost, etc. Will work on that as I have time if interested.

miloschwartz:
Hey thanks for working on this. This is one of those things that we cut down on at the beginning to release faster but we've always wanted. I will review this when I have some free time soon- hopefully this weekend or earlier.

grokdesigns:
@miloschwartz Glad to help! Quite a few changes in this most recent commit.

- Changed icon to more universal down-right arrow to indicate Invitations is a sub-menu of Users. Still only visible if invitation exist.
- Added the regenerate invitation functionality and have tested it with all combinations I can think of. This uses a modal popup like the Invite User function does, but does not allow changing the email address or user type, just the number of days the invitation is valid so it can be extended. Most of this is in the new RegenerateInvitationForm.tsx file.
- Replaced the method for keeping track of how many invites have been sent per hour in inviteUser.ts. Took the logic out of the initial Invite User functionality. The initial invite will only let you create one invitation for a given email address. Moved the new inviteTracker logic to the regeneration of invites. User will be able to create initial invite and then can regenerate up to 3 times within a one hour period. I changed your exiting logic to use node-cache instead of the previous solution where there was a comment of "// TODO: this could cause memory increase over time if the object is never deleted". I think node-cache will solve that, as it should clean up after itself. I haven't actually tested to ensure it is successfully reset after an hour, so should test that.
- Changed Invite User and Regenerate Invitation modals so that they always default to send email. Before, if you invited multiple users or regenerated multiple invitations, it would maintain the last state it was in. I think setting the default every time the modal opens is a cleaner user experience.
- Changed invitation email subject from "You're invited to join a Fossorial organization" to "You're invited to join {orgId}". Went back and forth on this one, but I think Fossorial might be confusing to end/less-technical users. Maybe this customization could be behind a subscription key?

Screenshots:
image
image
screenshot_20250409_20 51 53_32ibIt
screenshot_20250409_20 53 13_xqEtbD
screenshot_20250409_20 50 02_gTZVPf

grokdesigns:
@miloschwartz Glad to help! Quite a few changes in this most recent commit.

- Changed icon to more recognizable down-right arrow to indicate Invitations is a sub-menu of Users. Still only visible if invitations exist.

- Added the regenerate invitation functionality and have tested it with all combinations I can think of. This uses a modal popup like the Invite User function does, but does not allow changing the email address or user type, just the number of days the invitation is valid so it can be extended. Most of this is in the new RegenerateInvitationForm.tsx file.

- Replaced the method for keeping track of how many invites have been sent per hour to one user in inviteUser.ts. Took the logic out of the initial Invite User functionality. The initial invite will only let you create one invitation for a given email address regardless of time. Moved the new inviteTracker logic to the regeneration of invites. User will be able to create initial invite and then can regenerate up to 3 times within a one hour period. I changed your exiting logic to use node-cache instead of the previous solution where there was a comment of "// TODO: this could cause memory increase over time if the object is never deleted". I think node-cache will solve that, as it should clean up after itself. ~I haven't actually tested to ensure it is successfully reset after an hour, so should test that.~ Confirmed working.

- Changed Invite User and Regenerate Invitation modals so that they always default to send email. Before, if you invited multiple users or regenerated multiple invitations, it would maintain the last state it was in. I think setting the default every time the modal opens and making sure the form is consistent is a cleaner user experience and the default assumption is that an invite is sent and not sending an email would be the exception.

- Changed invitation email subject from "You're invited to join a Fossorial organization" to "You're invited to join {orgId}". I think Fossorial might be confusing to end/less-technical users? Maybe this customization could be behind a supporter key? Didn't change the signature in this commit, but also possibly something to change/enable with supporter key.


Screenshots:
image
image
screenshot_20250409_20 51 53_32ibIt
screenshot_20250409_20 53 13_xqEtbD
screenshot_20250409_20 50 02_gTZVPf

miloschwartz:
Thank you!

PR #501, opened 2025-04-09
Add option to customise TLS server name in resource settings

achtnullzwei:

## Description
Added an option to set a custom TLS Server Name to solve the issue with strict SNI checks. I have added it to the "General" page in Resource settings. Couldn't rly think of a better place, and didn't want to create a new section for it. Also it doesn't make sense to set this on target level. Happy to adjust if needed.

Usage:

- In general page of the resource settings, add the desired domain name to the new field TLS Server Name. Save. 
- To remove it, simply empty the field and save again.
- Value entered is checked against meaningful entries.

Value is stored in the resource table in a new column names tlsServerName. Hence, this would require to be covered in the migration scripts. I added a draft migration for future version 1.3.0 as a suggestion.

Screenshot:

432051205-ae8a5dea-bde4-4ffd-9f3f-b989e59af1e5

Notes: 
This is my very first contribution to any project on Github. I usually just work on my own project, but I felt I would like to contribute and get this feature implemented. In any case, I am kindly asking for some patience, if I did sth wrong. I am not rly used to write TS, but it was a lot of fun trying to understand the language a bit more, the code you created and add a feature to it. With respect to the pull request, I saw in the contribution guidelines, that contributions are supposed to be added to main branch. However, my changes would probably break without altering the database and having migration scripts properly added. I added a draft for that at least. Furthermore, I saw some recent pull requests made against dev branch. I hope that's ok. Otherwise, I can change it to the main branch. 

## How to test?

Build from sources and use docker image. I did test this as extensively as I could and can validate it works, at least in my homelab. I did test it on a live VPS with whoami containers and multiple levels of cascaded reverse-proxies. Also checked for UI issues as much as I can.

### DIFF
--- a/server/db/schemas/schema.ts
+++ b/server/db/schemas/schema.ts
@@ -78,7 +78,8 @@ export const resources = sqliteTable("resources", {
         .notNull()
         .default(false),
     enabled: integer("enabled", { mode: "boolean" }).notNull().default(true),
-    tlsServerName: text("tlsServerName").notNull().default("")
+    tlsServerName: text("tlsServerName").notNull().default(""),
+    setHostHeader: text("setHostHeader").notNull().default("")
 });
 
 export const targets = sqliteTable("targets", {
--- a/server/db/schemas/schema.ts
+++ b/server/db/schemas/schema.ts
@@ -77,7 +77,8 @@ export const resources = sqliteTable("resources", {
     applyRules: integer("applyRules", { mode: "boolean" })
         .notNull()
         .default(false),
-    enabled: integer("enabled", { mode: "boolean" }).notNull().default(true)
+    enabled: integer("enabled", { mode: "boolean" }).notNull().default(true),
+    tlsServerName: text("tlsServerName").notNull().default("")
 });
 
 export const targets = sqliteTable("targets", {
--- a/server/lib/schemas.ts
+++ b/server/lib/schemas.ts
@@ -9,3 +9,10 @@ export const subdomainSchema = z
     .min(1, "Subdomain must be at least 1 character long")
     .transform((val) => val.toLowerCase());
 
+export const tlsNameSchema = z
+    .string()
+    .regex(
+        /^(?!:\/\/)([a-zA-Z0-9-_]+\.)*[a-zA-Z0-9-_]+$|^$/,
+        "Invalid subdomain format"
+    )
+    .transform((val) => val.toLowerCase());
--- a/server/routers/resource/listResources.ts
+++ b/server/routers/resource/listResources.ts
@@ -68,7 +68,8 @@ function queryResources(
                 http: resources.http,
                 protocol: resources.protocol,
                 proxyPort: resources.proxyPort,
-                enabled: resources.enabled
+                enabled: resources.enabled,
+                tlsServerName: resources.tlsServerName
             })
             .from(resources)
             .leftJoin(sites, eq(resources.siteId, sites.siteId))
@@ -102,7 +103,8 @@ function queryResources(
                 http: resources.http,
                 protocol: resources.protocol,
                 proxyPort: resources.proxyPort,
-                enabled: resources.enabled
+                enabled: resources.enabled,
+                tlsServerName: resources.tlsServerName
             })
             .from(resources)
             .leftJoin(sites, eq(resources.siteId, sites.siteId))
--- a/server/routers/resource/listResources.ts
+++ b/server/routers/resource/listResources.ts
@@ -69,7 +69,8 @@ function queryResources(
                 protocol: resources.protocol,
                 proxyPort: resources.proxyPort,
                 enabled: resources.enabled,
-                tlsServerName: resources.tlsServerName
+                tlsServerName: resources.tlsServerName,
+                setHostHeader: resources.setHostHeader
             })
             .from(resources)
             .leftJoin(sites, eq(resources.siteId, sites.siteId))
@@ -104,7 +105,8 @@ function queryResources(
                 protocol: resources.protocol,
                 proxyPort: resources.proxyPort,
                 enabled: resources.enabled,
-                tlsServerName: resources.tlsServerName
+                tlsServerName: resources.tlsServerName,
+                setHostHeader: resources.setHostHeader
             })
             .from(resources)
             .leftJoin(sites, eq(resources.siteId, sites.siteId))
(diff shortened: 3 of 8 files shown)


achtnullzwei:
Not sure though how to best resolve the conflict thats listed.

miloschwartz:
Thanks @achtnullzwei! I will review this soon. I updated the docs to suggest opening the PR against the dev branch instead of main. Thanks for pointing that out.

achtnullzwei:
@miloschwartz Thx! I adjusted the field description slightly. I added "(optional)" to correctly indicate that this field is not required for all scenarios. I added the import in updateResources.ts to remove the conflict. Can be later merged again into one line again. There was also an issue in the proposal for the migration script, where the ALTER TABLE statement was wrong. I corrected this now, and tested with a 1.1.0 db backup. Worked in my homelab setup without an issue. 

Screenshot 2025-04-11 at 09 57 53

achtnullzwei:
@miloschwartz I have been thinking about it. Maybe its best to add an "Advanced" section. Reason is, at the moment we need to set "insecureSkipVerify" to true, hardcoded with my PR. Traefik will not merge serversTransport from HTTP provider with any default set in the file config. We can create a switch for it next to the new TLS Server name field in that section. We can default this to true in the DB to ease userexperience. Security wise, even Traefik suggests to set insecureSkipVerify on router level and not globally.

asychev:
@achtnullzwei Would it be better to make new option something like `targetHostName` instead of `tlsServerName`? Most of the time when target is behind another reverse proxy - you have to set both request SNI name and HTTP Host header to the same value another proxy expects.

achtnullzwei:
@asychev I checked how this is handled in CF tunnels. There it's actually 2 fields that need to be set. One field to set the SNI (TLS Server Name) and one field to rewrite the host header. I think this is done for good reason there. Setting the TLS Server Name helps if you do SNI based downstream routing, which in Traefik's case is done without terminating the TLS connection via TCP router. It just passes TLS through. Here host header rewrite wouldn't be relevant as Traefik is not inspecting the host header in such a scenario. It's a different story though if you have your pangolin resource set up on domain name A but downstream you need it to be rewritten to domain B. In such cases, both fields might be relevant and you need to terminate the TLS connection with a HTTP router. I think in this case, it might be best to add another field to rewrite host header as well. Tricky here is, I do not know if in a chain of middlewares, this needs to be the first middleware, or can be later in the chain. Would you happen to know at what step it's best to rewrite host header in the sequence of middlewares?

asychev:
@achtnullzwei Okay, that indeed makes sense (and also makes solution way too complicated). If we have to cover really all possible scenarios - having separate Host header and SNI name is necessary. Not sure most of Pangolin users need that, maybe my initial proposal is reasonable as first step. In my case I just can't use Immich via Pangolin as of today because it is behind NPM. I'm not familiar with Traefik in such deep details, unfortunately.

achtnullzwei:
@asychev you raised a very valid point though. I mean it gets complex, but that doesn't mean it needs to be for everyone. However, the flexibility would be needed I guess, to cover a wide range of scenarios. As I mentioned in one of my previous comments, maybe best to actually put this in an "Advanced" section, so its clear to users who don't require this complexity, that its optional and not needed for all scenarios. https://github.com/fosrl/pangolin/pull/511 suggests a changing the "SSL Configuration" section in "Connectivity" to "Advanced Configuration". If this change gets merged, it makes sense to put the TLS Server Name field, and possible "Set Host Header" field to the advanced section so it does not disturb the idea of a simple user experience. Furthermore, I would generally not rewrite a host header by default, or of its not needed. This would be hidden in the code of pangolin, so if you need to figure out where something is wrong in your chain of proxies, if you have a setup like this, it is not as transparent as having it as a separate setting.

oschwartz10612:
@achtnullzwei that's a good question. I will defer to @miloschwartz on this. I suggested moving the ssl out of the advanced section in #511 and moving it to the bottom of the page. Maybe it would make sense to put it in there?

achtnullzwei:
@miloschwartz @oschwartz10612 I have added another commit. To cover more scenarios, as per the discussion in the PR that was ongoing, I went to add a second field for host header rewrite as well. Additionally, to avoid cluttering the general section too much, I added an Advanced section to the general page that only shows for HTTP resources. This way it feels more optional to use. Adding this to "Connectivity" could be done as well, but since these changes are usually set just once for any connected target, I felt the General page is a better fit. This hopefully helps avoid any confusion. My last commit adds that.

Changes made: 
- Create section "Advanced" on General page for Resources (only shows for HTTP)
- Move TLS Server Name field to the new section
- Add field to rewrite hose header

This now solves the following scenarios:
- SNI based routing downstream
- Routing a.example.org downstream to b.example.org
- Setting a custom host header

I tested this with whoami docker containers. As far as I was able to test, adding the host rewrite at the end of the middleware array, does not affect if you have a secure headers middleware. test security headers test results were the same with or without header rewrite, so traefik merges this correctly between file and http provider. (To note: any additional host header rewrite via customRequestHeader.Host in the applied file middlewares should be removed. This might only be the case though for people tweaking the functionality with files due to missing features.)

Also tested if section displays properly, migrations run correctly, settings are set correctly in the DB, and in the traefik api endpoint. so far, I could not find any issue. Migration tested from 1.2.0 - the proposed 1.3.0 as in my PR.

This PR is based on the 1.2.0 main branch. Since then I think the dependencies have been updated. Not sure if there are any implications, couldn't test for this. 

Screenshot:
Screenshot 2025-04-15 at 13 27 51

Hope these adjustments make sense.

achtnullzwei:
<del> One last consideration for this. When setting a custom request header with the field, this only rewrites it downstream. It will not rewrite the response host header back to fullDomain. While this is not needed by default there can be scenarios where you want to do that. E.g when you route from domain a (pangolin) to domain b (target) and you dont want to expose domain b or the client application expects host header must be domain a. It should be rather simple to add this option via a switch. Let me know what you think and I can add it to the PR. </del>

asychev:
@achtnullzwei Not sure I understand what you talking about. There is no Host header in HTTP responses, can you show any examples?

achtnullzwei:
Youre right. I was mistaking this for sth else. I will remove the comment.

miloschwartz:
@achtnullzwei Thank you! Sorry this took so long for us to review.

achtnullzwei:
Thx a lot. No worries. Glad I could contribute a little 🙂

mschirrmeister:
What is the best way to test this?
I build an image from the `dev` branch and I can see the settings. It also says _successful_ when I hit save. But it does not seem to be actually saved. When I go out and back into the resource, the TLS field is again empty.

achtnullzwei:
Some commits happened in dev branch after my PR got merged that made some adjustment like formatting etc. Didnt do follow up tests since then. I can run a test tonight and check. But I assume all changes will be tested before release as its a quite active branch atm.

PR #511, opened 2025-04-11
Add ability for sticky sessions to backend resource.

x86txt:

## Description
Allow 'sticky sessions' to backend resources. This is required if there are multiple targets and the targets aren't stateless, like the VNC console sessions to Proxmox. This keep the user pinned to a single target for the duration of their session.

image

## How to test?
```shell
mkdir -p /tmp/server1 && echo "server1" > /tmp/server1/index.html && \
mkdir -p /tmp/server2 && echo "server2" > /tmp/server2/index.html && \
python3 -m http.server 10000 --bind 0.0.0.0 --directory /tmp/server1 && \
python3 -m http.server 20000 --bind 0.0.0.0 --directory /tmp/server2
```

Add two targets as shown in the screenshot below. Browse to your Resource and hit refresh and notice the webpage rotates from server1 to server 2, etc. Enable Sticky Sessions, hit refresh several more times, and notice that Traefik now keeps your session pinned to a single target.

image

### DIFF
--- a/server/routers/traefik/getTraefikConfig.ts
+++ b/server/routers/traefik/getTraefikConfig.ts
@@ -40,7 +40,8 @@ export async function traefikConfigProvider(
                     org: {
                         orgId: orgs.orgId
                     },
-                    enabled: resources.enabled
+                    enabled: resources.enabled,
+                    stickySession: resources.stickySession
                 })
                 .from(resources)
                 .innerJoin(sites, eq(sites.siteId, resources.siteId))
@@ -275,7 +276,18 @@ export async function traefikConfigProvider(
                                         url: `${target.method}://${ip}:${target.internalPort}`
                                     };
                                 }
-                            })
+                            }),
+                        ...(resource.stickySession
+                            ? {
+                                  sticky: {
+                                      cookie: {
+                                          name: "pangolin_sticky",
+                                          secure: resource.ssl,
+                                          httpOnly: true
+                                      }
+                                  }
+                              }
+                            : {})
                     }
                 };
             } else {
@@ -335,7 +347,18 @@ export async function traefikConfigProvider(
                                         address: `${ip}:${target.internalPort}`
                                     };
                                 }
-                            })
+                            }),
+                        ...(resource.stickySession
+                            ? {
+                                  sticky: {
+                                      cookie: {
+                                          name: "pangolin_sticky",
+                                          secure: resource.ssl,
+                                          httpOnly: true
+                                      }
+                                  }
+                              }
+                            : {})
                     }
                 };
             }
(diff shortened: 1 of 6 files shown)


oschwartz10612:
This is awesome thanks for tackling this! Some thoughts on the UI:

I think it might not make sense to lump enabling or disabling SSL in with the advanced options. Not sure how @miloschwartz might feel about that? Maybe it makes more sense to make a new section toward the bottom under the targets called advanced and leave ssl where it was? 

I am also thinking it might make sense to hide this option unless they have more than one target? Not committed to this one.

server/routers/traefik/getTraefikConfig.ts
```diff
@@ -335,7 +347,18 @@ export async function traefikConfigProvider(
...
+                            ? {
+                                  sticky: {
+                                      cookie: {
+                                          name: "pangolin_sticky",
+                                          secure: resource.ssl,
+                                          httpOnly: true
+                                      }
+                                  }
+                              }
+                            : {})
```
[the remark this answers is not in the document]
x86txt: Good catch! 🤜🏻💥🤛🏻 

Modified it to create a hash based off source IP + ephemeral port, which should account for multiple devices behind the same NAT. (server would see public IP + ephemeral port and sticky based on the hash of that). Will push a new PR shortly.

x86txt:
Love the comment about making it appear once a user inputs more than one target. I'm terrible at design though. I'll make it behave that way, stick it somewhere at the bottom, below the targets, which shouldn't be hard to position in a more visually pleasing location/manner.

server/routers/traefik/getTraefikConfig.ts
```diff
@@ -335,7 +350,17 @@ export async function traefikConfigProvider(
                                         address: `${ip}:${target.internalPort}`
                                     };
                                 }
-                            })
+                            }),
+                        ...(resource.stickySession
```
x86txt, line 354: Okay, added IP+PORT hash for TCP targets. This will use the full IP path, so it will account for source IP, XFF, etc. This should be able to handle multiple devices behind the same NAT well.

src/app/[orgId]/settings/resources/[resourceId]/connectivity/page.tsx
```diff
@@ -456,13 +486,24 @@ export default function ReverseProxyTargets(props: {
...
-                            SSL Configuration
+                            Advanced Configuration
                         </SettingsSectionTitle>
                         <SettingsSectionDescription>
-                            Set up SSL to secure your connections with certificates
+                            Configure advanced settings for your resource
                         </SettingsSectionDescription>
                     </SettingsSectionHeader>
                     <SettingsSectionBody>
+                        {targets.length >= 2 && (
```
Comment (x86txt) [inline src/app/[orgId]/settings/resources/[resourceId]/connectivity/page.tsx:496]: This will make the sticky sessions toggle visible only after the user enters 2 or more targets. I couldn't find a place where I thought it looked visually better, so please feel free to adjust the look if this PR is approved.

x86txt (commented):
Added two requested changes, please review.

oschwartz10612:
Getting to this soon we promise! Thanks for the hard work!

miloschwartz:
Thank you @x86txt! Sorry this took so long for us to merge.

PR #558, opened 2025-04-21
Added dependency of crowdsec to traefik and relaxed its health check

TuncTaylan:

## Description
Add dependancy on crowdsec for traefik and more relaxed health check for crowdsec, to prevent zombie traefik bouncers when starting the stack up.

Otherwise there are always "loose" or duplicate traefik bouncer plugins (check the last api pull dates):

```bash
docker exec -it crowdsec cscli bouncers list
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 Name                        IP Address  Valid  Last API pull         Type                             Version                                                                  Auth Type
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 traefik-bouncer             172.18.0.4  ✔️     2025-04-21T09:24:34Z  Crowdsec-Bouncer-Traefik-Plugin  1.X.X                                                                    api-key
 vps-firewall                172.18.0.1  ✔️     2025-04-21T09:32:16Z  crowdsec-firewall-bouncer        v0.0.31-debian-pragmatic-amd64-4b99c161b2c1837d76c5fa89e1df83803dfbcc87  api-key
 traefik-bouncer@172.18.0.3  172.18.0.3  ✔️     2025-04-18T20:43:20Z  Crowdsec-Bouncer-Traefik-Plugin  1.X.X                                                                    api-key
 traefik-bouncer@172.18.0.2  172.18.0.2  ✔️     2025-04-15T06:55:59Z  Crowdsec-Bouncer-Traefik-Plugin  1.X.X                                                                    api-key
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

## How to test?
Just run the installer and do a docker-compose.yaml comparison.

### DIFF
--- a/install/config/docker-compose.yml
+++ b/install/config/docker-compose.yml
@@ -48,10 +48,6 @@ services:
     depends_on:
       pangolin:
         condition: service_healthy
-{{if .DoCrowdsecInstall}}
-      crowdsec:
-        condition: service_healthy
-{{end}}
     command:
       - --configFile=/etc/traefik/traefik_config.yml
     volumes:
--- a/install/config/docker-compose.yml
+++ b/install/config/docker-compose.yml
@@ -48,6 +48,10 @@ services:
     depends_on:
       pangolin:
         condition: service_healthy
+{{if .DoCrowdsecInstall}}
+      crowdsec:
+        condition: service_healthy
+{{end}}
     command:
       - --configFile=/etc/traefik/traefik_config.yml
     volumes:
--- a/install/config/crowdsec/docker-compose.yml
+++ b/install/config/crowdsec/docker-compose.yml
@@ -9,6 +9,9 @@ services:
       PARSERS: crowdsecurity/whitelists
       ENROLL_TAGS: docker
     healthcheck:
+      interval: 10s
+      retries: 15
+      timeout: 10s
       test: ["CMD", "cscli", "capi", "status"]
     labels:
       - "traefik.enable=false" # Disable traefik for crowdsec
--- a/install/crowdsec.go
+++ b/install/crowdsec.go
@@ -3,9 +3,12 @@ package main
 import (
 	"bytes"
 	"fmt"
+	"log"
 	"os"
 	"os/exec"
 	"strings"
+
+	"gopkg.in/yaml.v3"
 )
 
 func installCrowdsec(config Config) error {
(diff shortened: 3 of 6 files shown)


install/config/docker-compose.yml
```diff
@@ -48,6 +48,10 @@ services:
     depends_on:
       pangolin:
         condition: service_healthy
+{{if .DoCrowdsecInstall}}
+      crowdsec:
+        condition: service_healthy
+{{end}}
```
oschwartz10612, line 54: This actually wont work because of the order of operations of the install and if you are doing an install after the fact once this .yml is already created.

oschwartz10612:
Thanks for the contribution!

We need to think a little bit about how to get the condition into the main .yml file. I would do it in a similar way I am updating the log volume in `CheckAndAddTraefikLogVolume` in the `config.go` file. I think we would need a function to check if the condition is there or not and only add it if its is needed and then call this function in the setup crowdsec workflow. Does that make sense?

install/config/docker-compose.yml
TuncTaylan: Yeah, sorry. I didn't invest much time to make sense, how the install routine worked.

TuncTaylan:
> Thanks for the contribution!
> 
> We need to think a little bit about how to get the condition into the main .yml file. I would do it in a similar way I am updating the log volume in `CheckAndAddTraefikLogVolume` in the `config.go` file. I think we would need a function to check if the condition is there or not and only add it if its is needed and then call this function in the setup crowdsec workflow. Does that make sense?

Yes, it makes absolutely sense. Is there any pseudo code or list of steps about how your installation routine works or what steps it needs to go through, in order to install or update?

I'm pretty sure, that it would be possible to do with Helm charts or Ansible scripts.

TuncTaylan:
> Thanks for the contribution!
> 
> We need to think a little bit about how to get the condition into the main .yml file. I would do it in a similar way I am updating the log volume in `CheckAndAddTraefikLogVolume` in the `config.go` file. I think we would need a function to check if the condition is there or not and only add it if its is needed and then call this function in the setup crowdsec workflow. Does that make sense?

Yes, that make sense. I'm now confident enough, that it works as intended.
I've refactored some of the existing recurring bits, added some error checking, cleared the existing warnings and implemented the function "CheckAndAddCrowdsecDependency".

Edit: Please excuse the harsh code, as it is my first experience with golang, my "hello, world!" into it ;)

oschwartz10612:
Thanks for the hard work! Looks good!

PR #598, opened 2025-04-25
Bug: fixed typo in exitNodes table

imbavirus:

## Description
The field publicKey is spelled wrongly in the exitNodes table

## How to test?

- Confirm spelling is correct in the database (exitNodes table)


oschwartz10612:
This is a very good catch thank you! 

I think changing the name of the column in the database will require a migration. We will handle that for the next release.

PR #699, opened 2025-05-10
fix: Avoid rendering two children with the same key

dbonne:
This change updates the value of the key 'Check ID' to 'checkId' to avoid duplicates. It ensures proper naming conventions in the permissions select box. Avoid rendering two children with the same key in the component.


miloschwartz:
Hey, thanks for bringing this to my attention! The permission ID needs to remain `checkOrgId` because that is what is used elsewhere, but I will remove the duplicate entries.

Issue #693, opened 2025-05-08
Traek access.log is growing and not log rotating.


LeonvanHeerden:
While going through all the folders in my running installation, I found that the `config/traefik/logs/access.log` was already over 2 gig and has never been log-rotated. 

* Restarting didn't make a difference. 
* I renamed the log, but since the file was still open with Traefik, it kept writing to the renamed file. 
* Then restarting, allowed a new file to be created. 
* The Pangolin logs in `config/logs/` are rotating as expected.

I checked inside the running containers', `/etc/logrotate*` configs for Traefik and Pangolin and neither has any configs that refers to logs. So am not sure how the rotate is currently working. 

What is the correct way to address this to properly create daily logs?

hhftechnology:
https://forum.hhf.technology/t/logrotate-configuration-guide-for-docker-services-in-pangolin-stack/

TuncTaylan:
A more appropriate solution would be to use the native log rotation instead of abruptly killing a running container, which would cause downtime.

Take a look here: [Traefik log rotation ](https://doc.traefik.io/traefik/observability/logs/#log-rotation)

I'm on the road at the moment, but I'll test a working configuration and get back to you. Maybe I'll extend the default configuration in the repository.

The important thing is not to delete or rotate logs that haven't been processed by Crowdsec, bearing in mind how fast the VPS/VM is and how quickly the logs can be parsed. As I said, I'll test this out with the least powerful VM possible and get back to you.

TuncTaylan:
@LeonvanHeerden , Check out this configuration, it works fine for me:
https://github.com/fosrl/pangolin/pull/736/commits/3ee4aaf19413ca8131f49c3422dfd1fc05acabd4

Here is the PR as well: https://github.com/fosrl/pangolin/pull/736

Edit: I now see, that it still needs an external intervention after getting a USR1 signal, there are some solutions for dockerized traefik, like https://github.com/vegardit/docker-traefik-logrotate.

TuncTaylan:
I'll evaluate the options and make further changes. One thing I would do is to do on the container level and not do adjustments on the host the stack run on.

TuncTaylan:
Here is a containerized solution using [vegardit/traefik-logrotate](https://github.com/vegardit/docker-traefik-logrotate):

[...]

PR #736, opened 2025-05-16
traefik log rotation with default values

TuncTaylan:

## Description
Enables traefik log rotation with default values that work with a least powerful VM.
The log file will get rotating after reaching 100MB of size and the last 3 backups are being kept and get compressed as well. The logs that are older then three days would get deleted.

This also should fix issues like https://github.com/fosrl/pangolin/issues/693


oschwartz10612:
Thanks great idea!

PR #740, opened 2025-05-17
integrating vegardit/traefik-logrotate

TuncTaylan:

## Description
Integrating a dockerized solution for traefik log rotation using [vegardit/traefik-logrotate](https://github.com/vegardit/docker-traefik-logrotate).

## How to test?
Just start the stack, it'll rotate the traefik logs.

### DIFF
--- a/docker-compose.example.yml
+++ b/docker-compose.example.yml
@@ -35,7 +35,7 @@ services:
       - 80:80 # Port for traefik because of the network_mode
 
   traefik:
-    image: traefik:v3.3.3
+    image: traefik:v3.4.0
     container_name: traefik
     restart: unless-stopped
     network_mode: service:gerbil # Ports appear on the gerbil service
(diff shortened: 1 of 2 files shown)


oschwartz10612:
Hi! 

Thanks so much for taking the time to put this together. I dont think we want to include by default an extra container right now. I think there was also basic log rotation added in #736.

Feel free to reopen and let me know your thoughts. I think this might be best as a docs page explaining how to add it if it really is doing something different enough to #736.

TuncTaylan:
Hey @oschwartz10612 ,

The PR [#736](https://github.com/fosrl/pangolin/pull/736) was intended to support log rotation and necessary configuration for Traefik. As outlined in [Traefik's documentation](https://doc.traefik.io/traefik/observability/access-logs/#log-rotation), log rotation still requires an external program to manage the actual rotation:

"Traefik will close and reopen its log files, assuming they're configured, on receipt of a USR1 signal. This allows the logs to be rotated and processed by an external program, such as logrotate."

The reasoning and context are described in [issue #693](https://github.com/fosrl/pangolin/issues/693), which is why I preferred a Dockerized approach over modifying and running external programs on the host.

Let me know your thoughts!

Edit: I can't reopen it without creating another PR, so I'll just wait for your answer.

buildplan:
This can be achieved by simply creating a simple log-rotation file in /etc/logrotate.d/ - like the below - Adding one more container is not something I would personally like just for this. 

```user@VPS:/etc/logrotate.d$ cat traefik 
/home/user/Pangolin/config/traefik/logs/access.log {
    daily
    rotate 5
    compress
    delaycompress
    missingok
    notifempty
    create 0644 user user
    postrotate
        # Send USR1 signal to the traefik container(id=7f1c8061f54c) to reopen logs
        docker kill -s USR1 7f1c8061f54c > /dev/null 2>&1
    endscript
}
```

TuncTaylan:
Of course, but in cases, where you aren't able or don't want to edit the host system, more secure and compatible approach is the dockerized way.

Also "logrotate" is not installed in every distribution...

oschwartz10612:
Ahh okay thanks for clarifying I was not aware of this I apologize. @miloschwartz what do you think? Not sure if we want this extra container always distributed or not in every install? I am tempted to suggest we make this a docs page?

oschwartz10612:
Hey sorry I have not given this the attention it deserves. @TuncTaylan what if we update the installer to include the logrotate when installing crowdsec and otherwise make sure the access log is turned off?

PR #796, opened 2025-05-29
Allow installer to run without requires `sudo`

socheatsok78:

## Description
Since we only need `sudo` to install Docker on system that doesn't have Docker installed. it doesn't make any sense to always required running the `installer` with `sudo`.

This PR changes that if you need the `installer` to only generate `docker-compose.yml` and configs, you don't need `sudo`. Unless Docker is no installed and you are running on `linux` based on the check in `main.go`:

image

## How to test?

N/A


oschwartz10612:
I think the only problem with this is that it creates the config files on line 70 before the docker install and then checks first thing if the config file is there. So if you did not use sudo then exited at the install question then reran it would think things are already installed. 

So we either need to improve that or change the order of installs. But I think the docker install might want to have the docker compose file.

socheatsok78:
~~Oh yeah! Let me move things around a little bit!~~

~~Hmmm, the `installCrowdsec` depends on `Docker` to be installed. The `config.TraefikBouncerKey` won't be able to get the api key since `crowdsec` container is not running.~~

~~I'll put this in to draft for now.~~

My goal is to avoid running the `installer` as `root` and I want to keep the change to a minimum. 

So let do it like this. I'll revert the changes and instead do a check if with `isDockerInstalled()` before checking if user is `root` or not.

oschwartz10612:
I think that should work! Let me test it now.

Thanks for the effort!

socheatsok78:
Awesome, let me know if any tweaks i need to make. Thank you!

oschwartz10612:
Oh only issue I can see now is the pulling of the containers. If docker is already installed but the user is not already in the docker group then it will fail. 

Maybe it should be a similar check? Or we can catch the permissions error when trying to do it and prompt to rerun or something. I think if you re run it will be the same error though so maybe we need to better check what stage we are at (if the containers are pulled) or something.  What do you think?

socheatsok78:
> Oh only issue I can see now is the pulling of the containers. If docker is already installed but the user is not already in the docker group then it will fail.

Yep, I'm aware of that. Let me see what i can do. Since I'm on `darwin`, i'm gonna have to spin up a vm so give me a moment.

oschwartz10612:
No rush thanks for taking a look at this!

socheatsok78:
This will perform a check if current user is in `docker` group, except on macOS. If the user is `root`, no check is needed.

oschwartz10612:
LGTM, thanks!

PR #817, opened 2025-06-01
docker run command in detached mode

PrtmPhlp:

## Description
Changed the `docker run` commands to run in detached mode, simplifying the workflow by allowing containers to run in the background without tying up the terminal.

## How to test?
---


oschwartz10612:
Thanks!

PR #855, opened 2025-06-06
fix - removed package-lock.json from .gitignore

Lokowitz:

## Description
Removed package-lock.json from .gitignore because it is not recommended.
Recreated package-lock.json to avoid errors with dependabot because dependabot is not working properly without.

https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5

I saw your merging conflict yesterday and i think this was the reason you added it to ignore?
I know the pain then updating deps and getting merging errors. Thats the reason i only update deps in the main branch and then merge it back to the other branches.

The problem you had was because i updated some deps in my PR against `dev` and you updated some deps at the same time in the `main`. If you still want to use `dev` as your main PR target i think you should update the deps directly into `dev` or into `main` and then merge it directly back to `dev`.

Have a nice day,
Marvin


oschwartz10612:
Yeah thanks for this. We were working through some install issues cross platform that I think package-lock was breaking. We are going to keep committing it. 

For the branches I think we will take your advice. We are new to dependabot and stuff so I appreciate the help. I will merge its PRs into the main branch and them we will merge down into dev etc...

On the last run I think upgrading some of the packages and the node version broke drizzle and we didnt have time to fix so I downgraded it. Need to check that out again. 

So you think we are good to leave the bot as it is on main and merge into dev? I dont think we want to make dev the default branch because we want people going to the more stable  main branch when the visit the repo.

PR #867, opened 2025-06-07
Fix - Sort Data In/Out correctly in Sites Management page in the Pangolin Dashboard

thijsvanloef:

## Description
Data Out and Data In sorting option in the "Manage Sites" tab does not actually seem to be sorting on most data in/out based on total number of bytes, but it will try to sort it as text (Str). 

The solution is to parse each size string to a numeric value (in bytes) and compare those numbers:

- Added a `parseDataSize` helper that splits off the numeric part and unit from the string and converts it to a consistent number (bytes)
- Added `sortingFn` to the `SitesTable.tsx` to correctly sort the data in/out.

As you can see the table now sorts correctly! 

image

## Fix Considerations
I've decided to put the `parseDataSize` function as a shared module, this makes it more reusable if the maintainers of this project decide to use any other table or component that needs to sort or compare data-size strings.
Adding it to only the `SitesTable.tsx` seemed a fine fix for now, but could lead to duplicate code if it is needed elsewhere in the project.

## How to test?
1. Clone the branch
2. Create a dev instance by following the https://docs.fossorial.io/development#pangolin docs.
3. Go to the sites overview in the pangolin dashboard.
4. Click on the sorting button above Data In or Data Out
5. Check if the Data In/Out is sorted on total number of bytes instead of as a string.

I'm not super used to Typescript, so please let me know if you would like things changed :)

### DIFF
--- a/src/app/[orgId]/settings/sites/SitesTable.tsx
+++ b/src/app/[orgId]/settings/sites/SitesTable.tsx
@@ -27,6 +27,7 @@ import { formatAxiosError } from "@app/lib/api";
 import { createApiClient } from "@app/lib/api";
 import { useEnvContext } from "@app/hooks/useEnvContext";
 import CreateSiteFormModal from "./CreateSiteModal";
+import { parseDataSize } from '@app/lib/dataSize';
 
 export type SiteRow = {
     id: number;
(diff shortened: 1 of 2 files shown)


oschwartz10612:
Thanks so much!

PR #871, opened 2025-06-08
feat - add test action

Lokowitz:

## Description
1. Added a simple test for PR and push on `main` and `dev`. The test is building and running the server and creates the docker file. If you have any other tests or test scripts, we can add it there.
2. Added package-lock.json to the docker build because i got multiple errors by building the docker image.
3. Added PHONY to `Makefile` because without it was not possible to run the `make test` command
4. Renamed `build-all` to `build-release` because the `make` command is `build-release`

Please double check!! It's my first github action test :)
Hope this helps with the dependabot updates to find the first big issues.

Greetings Marvin


oschwartz10612:
I think this would be good, thanks for all of your devops work on this project!

Two quick questions:

- Maybe we just do on PRs so we are not kicking off a long running runner every push to dev? Sometimes (maybe not kosher) we push frequently to dev. Whats standard?
- Should we be using `npm ci` in the Dockerfile instead of `npm install --only=production`?

Lokowitz:
> Maybe we just do on PRs so we are not kicking off a long running runner every push to dev? Sometimes (maybe not kosher) we push frequently to dev. Whats standard?

Yes you are right, thought that if it is just set to `on: pull_request` the check will just run once at PR creation and you need `on: push` for the updates on PRs but` on: pull_request` also include the updates, so i change it.

> Should we be using npm ci in the Dockerfile instead of npm install --only=production?

Was going through some docs and some big github projects and all i see is this:
`RUN npm ci --omit=dev && npm cache clean --force`
So npm ci is preferred because it is installing what is defined in `package.json` and `package-lock.json`.

Lokowitz:
Is there a reason why you don't copy these files from builder?
```
COPY server/db/names.json ./dist/names.json
COPY public ./public
```

oschwartz10612:
I dont think there was a reason that we dont copy those from the builder other than they are not really needed in the builder and only for running.

oschwartz10612:
Seems good to me! Thanks for the hard work as always!

PR #876, opened 2025-06-09
Define Files in eslint

thijsvanloef:

## Description
I was thinking about adding a eslinter to the github actions as well, however came across an error the existing `eslint.config.js`.
As there are no files defined in the `eslint.config.js` the linter will not actually run any of its checks on the typescript files.

If wanted, I can also add automated linting via github actions that run on every PR.

Feel free to close this PR if this is not needed/wanted.

## How to test?
1. Clone the repository
2. Run `npx eslint`
3. See that there are linting errors found:
    image

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -1,6 +1,7 @@
 // eslint.config.js
 export default [
     {
+        files: ["**/*.ts", "**/*.tsx", "**/*.js", "**/*.jsx"],
         rules: {
             semi: "error",
             "prefer-const": "error"


oschwartz10612:
I think this is great! I will accept the PR so we can use the linter thanks! 

I do think we should do a action that runs on PRs especially since Lokowitz opened a PR for testing. If you want to tackle that it would be awesome.

In terms of the problems we probably would want to fix those before we start running the action otherwise it does not do a lot of good because it is just one more error, right? The semicolons are easy and it looks like a lot of the rest of it is `@typescript-eslint/no-unused-vars`. We might have to put in some work to get that done.

thijsvanloef:
> In terms of the problems we probably would want to fix those before we start running the action otherwise it does not do a lot of good because it is just one more error, right?

Absolutely, if those errors are not fixed the check will just always fail, which is counter productive.

PR #888, opened 2025-06-11
Update packages

Lokowitz:

## Description
This is just to let you know, that i have tested some dep updates.
Included packe updates tested on my prod env, lookes good to me.

Regarding this [PR](https://github.com/fosrl/pangolin/pull/852):
Update of better-sqlite3 is not working for me.

There will come an update on tsx 4.19.4 to 4.20.1, this is throwing an ignored error on the org creation site after clicking create org.

Update to Node 22 seems to work, Node 24 still not working.

Feel free to create a docker image of my branch and test it.

Have a nice day
Marvin


oschwartz10612:
Yeah in #903 you can see the error `Error running setup functions: Transaction function cannot return a promise`. I think we might just need to go around and make the transactions not async but not 100% on that or if we even can easily...

PR #1023, opened 2025-07-06
Fix plural forms for Russian language

Burzhui34:
This PR changes plural forms from =1/other to one/other format in en-US.json.

This fixes grammar issues in Russian language:

Before: "2 дней" (incorrect)
After: "2 дня" (correct)
Need help: Could you help identify if there are other plural strings in the codebase that need similar changes?

This change should not affect English but will allow proper pluralization in Russian.


oschwartz10612:
Thanks for your continued hard work on the Russian translation!

PR #1040, opened 2025-07-09
Add podman support to the installer

wayneyaoo:

## Description
The change primarily adds support for podman installation within the installer. The installer will prompt for selection between docker (the default) and podman.

A few points to notice:
1. If podman and podman-compose are not available, the installer will exit. This is by design at the moment because installing for people like what the installer does to docker is involving. If this is needed, I can do that in later iterations.
2. I added `docker.io` to the compose template file. This is because for freshly installed podman, it needs to add an entry to the `registeries.conf` so that image short name resolves. But I don't think editing this on behalf of people makes a lot of sense, so I resort to use explicit full name in the image entry. This has benefit to docker too because explicitness is still better than being implicit.
3. I added a few targets to the Makefile without breaking existing ones in CI. This is easier for local dev.

## How to test?
Spin up a Debian/Ubtuntu machine and manually install podman and podman-compose, then run the installer.


wayneyaoo:
Note to maintainers/reviewers:

I noticed the installer is mostly written specifically for docker in a procedural way. With the new support of `podman`, I'm happy to add a thin layer of abstraction to handle both podman and docker operations behind a maybe `IContainer` to make the code easier to follow. But I'd like to understand what people think on the style.

Also I noticed there's no unit test for the installer. Not sure if there're integration tests either. So if maintainers are good with interface-based mocks for UT, I can do a bit of rehaul to the installer in general for better test coverage. Let me know.

wayneyaoo:
Will use another local feature branch to do PR. That makes future commit easier.

wayneyaoo:
See https://github.com/fosrl/pangolin/pull/1041

PR #1041, opened 2025-07-09
Add podman support to the installer

wayneyaoo:

## Description
The change primarily adds support for podman installation within the installer. The installer will prompt for selection between docker (the default) and podman.

A few points to notice:
1. If podman and podman-compose are not available, the installer will exit. This is by design at the moment because installing for people like what the installer does to docker is involving. If this is needed, I can do that in later iterations.
2. I added `docker.io` to the compose template file. This is because for freshly installed podman, it needs to add an entry to the `registeries.conf` so that image short name resolves. But I don't think editing this on behalf of people makes a lot of sense, so I resort to use explicit full name in the image entry. This has benefit to docker too because explicitness is still better than being implicit.
3. I added a few targets to the Makefile without breaking existing ones in CI. This is easier for local dev.

## How to test?
Spin up a Debian/Ubtuntu machine and manually install podman and podman-compose, then run the installer.

For some reason, Debian 12 default podman and podman-compose's versions are too old and they have problem processing the `docker-compose.yml`. I believe this is version-specific, not distro-specific. We should specify to use newer versions (>= v5.5.2 of podman, >= 1.4.1 podman-compose). Otherwise it'll appear as pangolin installation being broken which is not the case.

### DIFF
--- a/install/main.go
+++ b/install/main.go
@@ -79,7 +79,7 @@ func main() {
 			os.Exit(1)
 		}
 
-		if err := exec.Command("bash", "-c", "cat /etc/sysctl.conf | grep 'net.ipv4.ip_unprivileged_port_start='"); err == nil {
+		if err := exec.Command("bash", "-c", "cat /etc/sysctl.conf | grep 'net.ipv4.ip_unprivileged_port_start='").Run(); err != nil {
 			fmt.Println("Would you like to configure ports >= 80 as unprivileged ports? This enables podman containers to listen on low-range ports.")
 			fmt.Println("Pangolin will experience startup issues if this is not configured, because it needs to listen on port 80/443 by default.")
 			approved := readBool(reader, "The installer is about to execute \"echo 'net.ipv4.ip_unprivileged_port_start=80' >> /etc/sysctl.conf && sysctl -p\". Approve?", true)
@@ -389,7 +389,6 @@ func createConfigFiles(config Config) error {
 
 		return nil
 	})
-
 	if err != nil {
 		return fmt.Errorf("error walking config files: %v", err)
 	}


wayneyaoo:
Note to maintainers/reviewers:

I noticed the installer is mostly written specifically for docker in a procedural way. With the new support of `podman`, I'm happy to add a thin layer of abstraction to handle both podman and docker operations behind a maybe `IContainer` to make the code easier to follow. But I'd like to understand what people think on the style.

Also I noticed there's no unit test for the installer. Not sure if there're integration tests either. So if maintainers are good with interface-based mocks for UT, I can do a bit of rehaul to the installer in general for better test coverage. Let me know.

oschwartz10612:
@wayneyaoo this is great thank you! To respond to some of your thoughts: 

- Yes I think an abstraction layer for the container orchestration layer would be nice if we are going to support multiple going forward. I will never say no to better code! : ) 
- Unit and integration tests would be amazing as well! I will never say no to tests! : )

I can give it a test in a bit and see how it works out. If its looking good do you want to merge this and then take a look at these proposed changes or do you want to open a new PR?

wayneyaoo:
@oschwartz10612 Thank you for being open-minded :) I think at the moment if we merge it first (provided that the functionality is ok) and get some feedback from people (if any), that'll help for future improvement. In the next PR I can do a rehaul to abstract stuff and add UT. How's that?

wayneyaoo:
One more question: what is your thought on installation with CLI args:)? I mean pure interactive installation experience is fine but sometimes supplying values via CLI args with interactive asks as fallback seems more flexible. I know this probably isn't a big deal - people won't install pangolin all day lol. So just asking.

oschwartz10612:
Great sounds like a plan. I will give it a test ASAP and then merge if it looks good. 

CLI args would be good too with the fallback to interactive if not provided as you say. Maybe one day that opens up the ability to make a nice web configuration that spits out a command or something for people to run.

wayneyaoo:
:) gentle ping

oschwartz10612:
Hey @wayneyaoo yeah I tried it and it was not working on ubuntu either. I think the containers started but I could not access the UI at all. I had: 

```
podman-compose version 1.5.0
podman version 4.9.3
```

from installing it using `apt install podman`

Is this what you were seeing? How are you installing podman >5 on Ubuntu/Debian? Do I need to build from source? 

Edit: Never used podman before lol sorry

wayneyaoo:
Interesting what is the error? I suspect distro-packed podman version is old. Maybe try at least podman version > 5.0? Podman compose version seems fine.

Also I noticed you tried to merge master by pushing a few commits - I didn't have a lot of PRs on Github before but is this normally how it's get done? Wouldn't it conflict the master later?

oschwartz10612:
Sorry for the delay. At the  containers seemed to go up fine and run but I was not able to access the web UI or anything externally. Using docker on the same install worked fine. I was using sudo so I dont think the port range thing was getting me?

I was wondering if it could be that the older versions dont support the `service:gerbil` network config or something and thats why the ports would not be exposed correctly?

wayneyaoo:
I did found a bug somehow escaped the manual testing - but you're correct that running as root shouldn't have access issue. I fixed the bug and push a new commit.

But interestingly I couldn't get 1.7.3 working. The Pangolin container seems to be pending on start and stuck there forever. Could you confirm 1.7.x is a working release? I couldn't manage to test my change with the new release even though I downgraded to 1.6.2. (but it could be my podman installation too. Didn't realize how unfriendly it is to install a specific version of podman and `podman compose`... I'm new to podman too)

If 1.7.x confirms to work could you test the change again please? I remove those health check locally and the dashboard ends up with a 404 not found and Pangolin at 3001 not available - but at least traefik is up.

oschwartz10612:
Realized maybe docker was messing with podman so made sure to stop docker when I was testing. Seems like it worked for me! I was able to run the installer with just the basic install of podman on ubuntu 24.04 and it seems to work fine!

Maybe the only thing I would consider is making podman a CLI arg instead of a question until we have the install built in? I think its just one more step most people dont want to think about but if you do want to use podman you can always use `--podman`.

Otherwise ready to merge I guess!

wayneyaoo:
Are you suggesting in this PR we make the 1st question a CLI arg until we have `podman` auto installation like `docker` does? I'm good with that but given there's no CLI argument defined at the moment, making one there with a temporary purpose sounds ... inconsistent from future perspective. If we can get this in first, I'll prioritise a rewrite with CLI args. That should work in long term withouit a middle state. How's that?

oschwartz10612:
Sounds good to me!

PR #1087, opened 2025-07-18
Small Typo causes crowdsec to fail

itsbhanusharma:
first rule is named iame instead of name. seems like a recent typo. I edited file manually and it seems to have allowed crowdsec to boot up.


oschwartz10612:
Thank you!

PR #1089, opened 2025-07-18
test if smtp user/pass config is set and if not set auth: null

tomribbens:

## Description
For smtp servers that require no authentication (Google Workspace has this option for whitelisted IP addresses for example), the `auth:` option for nodemailer must be set to null. This code checks if a username and password are defined. If they are, there is no change to previous behaviour. But if they aren't, `auth:` is set to null.

As discussed on discord: https://discord.com/channels/1325658630518865980/1395139357312028672

## How to test?
Don't set smtp user/pass values in the config, and configure an open relay as smtp host, and see if it mails.


oschwartz10612:
Thanks! Appreciate it!

PR #1104, opened 2025-07-21
add nixos option for newt in site creation

jackrosenberg:
Hello, 

I have added instructions for running Newt as a service [via NixOs](https://github.com/NixOS/nixpkgs/pull/414337), for which I am a maintainer. [Pangolin](https://github.com/NixOS/nixpkgs/pull/416148) is waiting to be merged too. 

## Description

Adds NixOs to the list of install options. 

## How to test?

Follow normal installation steps, click on the NixOs icon in the site creation process :)


src/app/[orgId]/settings/sites/create/page.tsx
```diff
@@ -285,6 +288,22 @@ WantedBy=default.target`
...
+NEWT_ID=${id}
+NEWT_SECRET=${secret}
+
+Enable the NixOs service for Newt, by adding the following to your configuration.nix:
+
+services.newt = {
+  enable = true;
+  environmentFile = /path/to/file_created_above.env;
+};`
+                ]
```
Comment (SigmaSquadron) [inline src/app/[orgId]/settings/sites/create/page.tsx:306]:
```suggestion
                ],
                aarch64: [
                        `# Make a new file, owned by 'root' and with mode '700',
# containing your newt secret as an environment variable:

echo "NEWT_SECRET=${secret}" > /var/lib/secrets/newt
chmod 700 /var/lib/secrets/newt
chown root:root /var/lib/secrets/newt

# Enable the NixOS service for Newt, by adding
# the following to your configuration.nix:

# services.newt = {
#   enable = true;
#   endpoint = "${endpoint}";
#   id = "${id}"; # You can also hide your Newt ID from your configuration by adding the 'NEWT_ID' variable to the environment variable file.
#   environmentFile = /var/lib/secrets/newt;
# };

# Then, rebuild your system:

nixos-rebuild switch`
                ]
```

src/app/[orgId]/settings/sites/create/page.tsx
```diff
@@ -285,6 +288,22 @@ WantedBy=default.target`
                 "Podman Run": [
                     `podman run -dit docker.io/fosrl/newt --id ${id} --secret ${secret} --endpoint ${endpoint}`
                 ]
+            },
+            nixos: {
+                x64: [
```
Comment (SigmaSquadron) [inline src/app/[orgId]/settings/sites/create/page.tsx:293]:
```suggestion
                x86_64: [
```

oschwartz10612:
I dont think those eslint errors are related to this PR. Can fix independently.

oschwartz10612:
Thanks for the PR! I do think we want to get Nixos into the UI! 

Giving this a think I am not sure if we want all of that instruction in the UI itself. I think it makes it big and we are not explaining to people how to run docker for example.

I am not super familiar with nixos, but could we par it down to just what we need? Maybe for example just the section you need to put into the file.  Then we can add a link in the UI to a docs page where the rest is written up. Could that work? Creating the files and stuff would go into the doc writeup

jackrosenberg:
How's this @oschwartz10612 ?

src/app/[orgId]/settings/sites/create/page.tsx
```diff
@@ -285,6 +288,14 @@ WantedBy=default.target`
                 "Podman Run": [
                     `podman run -dit docker.io/fosrl/newt --id ${id} --secret ${secret} --endpoint ${endpoint}`
                 ]
+            },
+            nixos: {
```
Comment (SigmaSquadron) [inline src/app/[orgId]/settings/sites/create/page.tsx:292]: Instead of NixOS, this command can run on pretty much any machine with Nix installed.
I'm not sure how to best expose that this command can be used in any arch or OS that has Nix.

oschwartz10612:
Seems good. Looks like we are using the command `nix run 'nixpkgs#fosrl-newt' --id ${id} --secret ${secret} --endpoint ${endpoint}`!

Thanks for the PR!

Issue #110, opened 2025-01-30
X-Forwarded-For: not working with IPv6


danhusan:
It seems X-Forwarded-For and X-Real-Ip shows the IP of badger when resources are accessed by IPv6.
Beta8

wget -4 whoami.mydomain.com
...
...
X-Forwarded-For: 193.193.221.3
X-Real-Ip: 193.193.221.3
...
...

wget -6 whoami.mydomain.com
...
...
X-Forwarded-For: 172.18.0.1
X-Real-Ip: 172.18.0.1
...
...

oschwartz10612:
Thanks for this. I have not focused on IPv6 support but it is something I want to do more of at some point.

I can look into this - I assume it has to do with some v4-v6 translation stuff. Is this impacting any functionality for you right now?

danhusan:
Does not have any impact for me atm, just trying to document things that can be improved in the future.

zandercodes:
To obtain an IPv6 forwarding address, you must be activated in Docker and on your server.

Add this to the Traefik Container in the Docker Compose or you can also configure in the sysctl on the Linux System
```yaml
sysctls:
  net.ipv6.conf.all.disable_ipv6: 0
  net.ipv6.conf.all.forwarding: 1
```

Add this to the Docker Daemon config (Example):
```json
{
    "ipv6": true,
    "experimental": true,
    "ip6tables": true,
    "fixed-cidr-v6": "fd07:a:b:c::/64",
    ...
}
```
IPv6 is an experimental feature in Docker

4ndreasH:
I have a Pangolin system with the plugins Crowdsec, Geoblock, and Umami-feeder (see #914 for details). These three plugins require the real IP address to function correctly. Everything works fine with IPv4, but for IPv6 they seem to be bypassed (probably due to the missing X-Forwarded-For header). 

Geoblock:
I have analyzed the log entries from Geoblock over the past five days, and not a single IPv6 address appeared.
(For comparison, I have a server using a Traefik proxy (without Pangolin), fed by HAProxy using proxy-protocoll. On that system, I do see blocked IPv6 addresses every day.)

Crowdsec:
To test crowdsec I added my external IP address to the decision list and tried to access my homepage. With IPv4 I get 403 (forbidden) as expected, but for IPv6 I could access my homepage without any restrictions.

Umami:
Accessing my web pages via IPv6 works, but those accesses are not being recorded by the Umami-feeder.

[...]

Issue #1084, opened 2025-07-18
1.7.1 | Client IP is reported incorrectly


elcajon:
After updating to version 1.7.0 or 1.7.1, I also initially noticed the behaviour described in issue #1081. Some services work, but others do not. However, I have noticed something else that I am not sure is related to this or another issue.

Since updating Pangolin, there has been an issue with the X-Forwarded-For header in my Pocket-ID access log: only the internal Docker network IP is reported.

<img width="1269" height="1106" alt="Image" src="https://github.com/user-attachments/assets/c4316a07-19c7-44d5-9e45-940bf97d5337" />

R-Nabil:
So I do have the issue reported in #1081 but I do not have your issue. 

I do use pocket id but my IP is reported appropriately.

Xentrice:
Is your client using IPv6 by any chance?
See #110 - this might be related to docker rather than the update.

elcajon:
> Is your client using IPv6 by any chance?
See https://github.com/fosrl/pangolin/issues/110 - this might be related to docker rather than the update.

Nice catch, thanks for that. I realized all the connections made today have been IPv6 instead of my home network which relies on IPv4. So I will close this. Thanks again.

PR #1111, opened 2025-07-22
add IPv6 support for docker network

Xentrice:

## Description
Adding IPv6 support to the docker network to ensure no NATing takes place and the IP addresses are logged correctly. This ensures correct behavior of middlewares, for example Crowdsec or GeoBlock.
Successfully tested on Debian based systems with different IP configurations (IPv4 only/IPv6 only/both). Also successfully tested on an existing install (IPv4/6 dual, Debian 12).
I did however only test on Hetzner VPS. There might be some differences in the server setup across providers, but i doubt they would impact this change.

Some follow up thoughts regarding #110:
Upon further thinking I came to the conclusion that step 1 might be disruptive - blocking IPv6 completely might break some installs if users are using an IPv6 only server (unlikely, as GitHub doesn't have IPv6 support yet, but not completely ruled out).

I also put some thoughts again into enabling IPv6 by default on all new installations, and can't think of any reason not to. My initial hesitation was due to increased complexity, especially with dockers tendency to punch through existing firewall policies. After giving this some thought, i came to the conclusion that some basic tech knowledge can be expected from the user, and I think it is a reasonable expectation that someone provisioning an IPv6 capable server is aware of the implications. So, as of now, I don't see any downside to enable IPv6 on all new installations.

In the end, this was basically just adding one line and running a lot of tests. So @oschwartz10612... https://github.com/fosrl/pangolin/issues/110#issuecomment-2994275283 you were completely correct :)

Changing this to ask the user during installation would be trivial now that the tests are done, just let me know if you prefer this. TBH tho, after my tests, I don't see a reason anymore.

I am unsure about how to handle existing installations, if at all. A note in the release notes might be sufficient.

## How to test?
Provision a fresh install or change the docker-compose.yml on an existing install
Ensure connection works from both IPv4 and IPv6 clients
Ensure the IP forwarding works as expected, for example by checking the traefik access.log. IPv6 addresses should now show correctly.

### DIFF
--- a/docker-compose.example.yml
+++ b/docker-compose.example.yml
@@ -52,3 +52,4 @@ networks:
   default:
     driver: bridge
     name: pangolin
+    enable_ipv6: true
--- a/install/config/docker-compose.yml
+++ b/install/config/docker-compose.yml
@@ -59,3 +59,4 @@ networks:
   default:
     driver: bridge
     name: pangolin
+    enable_ipv6: true


oschwartz10612:
Tested with V6 and did not see any issues. Thinking this is probably fine? 

I am leaning toward maybe making it an option during install in case it does break anything for people's installs. That way they had the option to remove it if they wanted to but it can default to yes as I agree there does not appear to be too much downside. Do you agree? Would you be able to make it conditional in the installer? 

If not I can.

Xentrice:
Hm, the thing is, if you disable this during a default install and your server is generally IPv6 capable, you don't disable IPv6, you just enable NATing, meaning every IPv6 connection gets translated to the docker container internal IPv4. This is  the reason for the initial issue.
I'm not sure if it's reasonable to explain this during the installation prompts without making them too bloated.

If the prompt should actually disable IPv6, some additional changes are needed - imho it would be sufficient to publish the gerbil docker ports to IPv4 only.

I do admit I lean towards it just being the default, after all IPv6 is inevitable and during all my tests i didn't see a single issue. Maybe a two step progress is still the way though, enabling it with opt-out, just in case there are some niche cases i didn't think of or i missed something, and somewhere down the line making it default.

I can make all the changes needed including the conditionals, just tell me which way to go.

oschwartz10612:
Good reasoning I think I agree! Will merge!

Thanks for doing all of this testing and giving this some thought. Appreciate it!

elcajon:
I became aware of this based on my issue #1084. FYI -  Having manually adjusted my configuration, I can confirm that it works well for me in a productive environment. Thanks a lot for the contribution- it's really appreciated! 😊 

image

PR #1132, opened 2025-07-26
Form Sign-in Improvements - Updated Login Page with Tabbed Interface

adrianeastles:

## Description
- Add tabbed interface for password, passkey, and SSO login
- Prioritize SSO as first option when OIDC providers are configured
- Improve UX with cleaner authentication methods.

## How to test?
Screenshot 2025-07-26 at 12 01 54 am

Screenshot 2025-07-26 at 12 02 20 am

Screenshot 2025-07-26 at 12 02 25 am

### DIFF
--- a/src/components/LoginForm.tsx
+++ b/src/components/LoginForm.tsx
@@ -26,7 +26,7 @@ import { LoginResponse } from "@server/routers/auth";
 import { useRouter } from "next/navigation";
 import { AxiosResponse } from "axios";
 import { formatAxiosError } from "@app/lib/api";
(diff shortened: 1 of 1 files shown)


oschwartz10612:
Hey @adrianeastles sorry for the delay on this. We really appreciate the time you put into it. 

When testing we realized there are two clicks required if you are not logging in with the initially selected option. So for example if you always log in with the passkey now you have to click passkey and then log in with passkey. Two clicks instead of one. We felt like the current method actually is less friction for the user.

shanelord01:
Can we get the option to disable specific login options (disable username/password) ?

PR #1138, opened 2025-07-27
add an environment variable for the smtp_pass config option

SigmaSquadron:

## Description
The password for secure authentication may be sensitive, so it is best to not leave it lying around in a config file. This commit introduces the `EMAIL_SMTP_PASS` environment variable, which can be set to configure the SMTP password without writing it to the configuration file.

## How to test?
Start a Pangolin server as normal, but use `EMAIL_SMTP_PASS` instead of the config file to set up the mail server.


oschwartz10612:
Thanks!

PR #1139, opened 2025-07-27
add shebangs to migration and server scripts

SigmaSquadron:

## Description
In NixOS, we wrap these files in a bash script to allow users to just run them as normal executables, instead of calling them as arguments to Node.JS. In our build scripts, we just [add the shebang](https://github.com/NixOS/nixpkgs/blob/d299df66b63ba4e6f2f98f5b1e13bedd7b748d8b/pkgs/by-name/fo/fosrl-pangolin/package.nix#L100) after the files have been compiled, but adding it upstream will allow all Pangolin users to just run ./server.mjs to start their Pangolin instances.

## How to test?

Build Pangolin, and check if all `*.mjs` files have shebangs. (and not just `cli.mjs`)

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -1,3 +1,4 @@
+#! /usr/bin/env node
 import "./extendZod.ts";
 
 import { runSetupFunctions } from "./setup";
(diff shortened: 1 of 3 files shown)


oschwartz10612:
Thanks!

PR #1175, opened 2025-07-31
Fix to issues with binding to ports other than 80/443 for web resources

Error-Gap:

## Description
Fix to issues with binding to ports other than 80/443 for HTTP/HTTPS web resources

- server/routers/badger/verifySession.ts : verifyResourceSession() updated code behind "cleanHost" var to a regex which strips the trailing :port for any port (rather than a string match for 80/443)
- src/app/auth/resource/[resourceId]/page.tsx : ResourceAuthPage() added a secondary match for serverResourceHost and redirectHost that accounts for ports
- server/routers/badger/exchangeSession.ts :  Updated exchangeSession() to use the same "cleanHost" type var (with port-stripping) as in verifyResourceSession(), replaced references to "host" with "cleanHost"

## How to test?

When running from docker, bind the ports in docker-compose with something other than 443:443 and 80:80
(in my case I also changed the "entrypoints" ports for traefik but this may not be necessary)

### DIFF
--- a/server/routers/badger/exchangeSession.ts
+++ b/server/routers/badger/exchangeSession.ts
@@ -53,19 +53,26 @@ export async function exchangeSession(
     try {
         const { requestToken, host, requestIp } = parsedBody.data;
 
+        let cleanHost = host;
+        // if the host ends with :port
+        if (cleanHost.match(/:[0-9]{1,5}$/)) {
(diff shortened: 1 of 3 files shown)


Error-Gap:
Wrong branch. Will re-submit for dev

PR #1177, opened 2025-07-31
Portbinding fixes

Error-Gap:

## Description
server/routers/badger/verifySession.ts : verifyResourceSession() updated code behind "cleanHost" var to a regex which strips the trailing :port for any port (rather than a string match for 80/443)
src/app/auth/resource/[resourceId]/page.tsx : ResourceAuthPage() added a secondary match for serverResourceHost and redirectHost that accounts for ports
server/routers/badger/exchangeSession.ts : Updated exchangeSession() to use the same "cleanHost" type var (with port-stripping) as in verifyResourceSession(), replaced references to "host" with "cleanHost"

## How to test?
When running from docker, bind the ports in docker-compose with something other than 443:443 and 80:80
(in my case I also changed the "entrypoints" ports for traefik but this may not be necessary)


oschwartz10612:
Awesome thanks for the fix this is great! Curious if newt has an issue with connecting when this is the case? Have you tested with newt?

Error-Gap:
Well damn. Now that I think about it: no. I've been using WireGuard for all my connections.

I can probably throw up a test VM and Newt agent today or tomorrow to see if any issues crop up with that.

The files/modules being mentioned in Lint don't look like any I've actually edited though, so the semicolon errors etc may be unrelated

oschwartz10612:
Yeah ignore the lint and build - might be something from dev sorry about that. I am also working on fixing the lint errors. 

That would be great if you could test newt. My hypothesis is that it will not like having a port in the endpoint but you never know. 

If I have a sec I can try to take a look but you might be able to get to it first lol.

Error-Gap:
Did a quick test with Newt and it seems to work, so long as the alternate port is configured on the dashboard_url in config.yml

oschwartz10612:
Awesome thanks so much!

Issue #1131, opened 2025-07-25
[Feature Request] System wide default language


al-dubois:
Is there a way to set the default language to something other than english for all users ? I support a french speaking company and they don't like that the default language isn't french.

Thanks 👍

softcat:
I totally agree — having the option to set a default language other than English would be really useful. I'm supporting users in a similar situation, and many of them expect to see their native language right from the start.

al-dubois:
I think we could try to use the browser's language property https://developer.mozilla.org/en-US/docs/Web/API/Navigator/language and fallback to english.

Thoughts ?

oschwartz10612:
Should be an easy add. Pulling from the browser will be the easiest first thing to do!

Pallavikumarimdb:
@oschwartz10612 , I want to work on this. Can you assign it to me?

oschwartz10612:
@Pallavikumarimdb that's awesome!

Pallavikumarimdb:
@oschwartz10612 , Please review the PR and let me know if any changes are needed.

al-dubois:
Can't wait to see it in prod ;)

Thanks @Pallavikumarimdb

oschwartz10612:
Next release ;}

PR #1205, opened 2025-08-03
System wide default language detection via browser language header

Pallavikumarimdb:

## Description
This PR adds support for detecting a user's preferred language from their browser settings (Accept-Language header) and falling back to a predefined default (en-US) when not available. The flow works as follows:

1. If a cookie exists and its value matches one of the app-supported locales (/i18n/config.ts), it is used.
2. If not, the Accept-Language header from the browser is parsed.
        We extract the base language (e.g., fr from fr-FR).
        We try to match it against available locales by comparing prefixes.

3. If no match is found, the app falls back to the default locale: 'en-US'.

## How to test?
1. Open your browser settings and change the preferred language to something like French.
2. Make sure the NEXT_LOCALE cookie is not set by clearing cookies or using an incognito/private window.
3. Visit the app in a new tab or incognito window.
4. Set your browser to a language not included in the app’s locales. The application should fallback to en-US.
5. Cookie Override Test: If NEXT_LOCALE cookie is set (e.g., via language switcher), it should take priority over browser settings on the next visit.

Fix: #1131

### DIFF
--- a/src/services/locale.ts
+++ b/src/services/locale.ts
@@ -1,7 +1,7 @@
 'use server';
 
-import {cookies, headers} from 'next/headers';
-import {Locale, defaultLocale, locales} from '@/i18n/config';
+import { cookies, headers } from 'next/headers';
+import { Locale, defaultLocale, locales } from '@/i18n/config';
 
 // In this example the locale is read from a cookie. You could alternatively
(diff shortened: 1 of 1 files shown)


oschwartz10612:
Thanks so much this is awesome!

PR #1208, opened 2025-08-03
Add setup token security for initial server setup

adrianeastles:
### Problem
When spinning up a new Pangolin server, anyone in the world could potentially access the initial setup page and create the first admin account before the legitimate administrator, posing a security risk.

### Solution
Implemented a setup token system that requires a secure token to be entered during the initial server setup process. The token is generated and displayed in the server console on startup until the token is used.

### **Initial Setup Page**
1.  Navigate to `http://localhost:3002/auth/initial-setup`.
2.  You should see a new "Setup Token" field at the top.
3.  Enter the token from the server console.
4.  Fill in the email and password fields.
5.  Submit the form.
6.  **Expected Behavior:**
    * ✅ The form accepts a valid token and creates the admin.
    * ❌ The form rejects an invalid token with an error message.

### Screenshots

Screenshot 2025-08-03 at 9 25 33 pm

Screenshot 2025-08-03 at 9 27 23 pm

Screenshot 2025-08-03 at 9 27 46 pm

Screenshot 2025-08-03 at 9 28 15 pm


adrianeastles:
Hey @miloschwartz, I've created this pull request to add a token requirement for the initial user registration. This token is generated on startup and is used by the user in the frontend to validate the initial user with the backend. This simple implementation helps prevent unauthorized access by adding one more extra setup before creating the main admin account.

oschwartz10612:
Hey @adrianeastles thanks for adding this! Definitely needed.  

Only thing I think this needs is some sort of integration with the installer. People will get confused once they install and there is no clear way to see the token. We might need to have the installer grab the Docker logs and dump it to the terminal for them to see AND/OR look for the token in the logs and print it to the screen. Or some other integration like the installer actually inserts the token into the sqlite DB and then print it out. Does that make sense?

adrianeastles:
Yeah makes sense, I'll see what I can do tonight and loop back.

adrianeastles:
@oschwartz10612 I've added the code into the installer script that will pull out token from the logs and let the user know. Again it's not 10/10, but it get's the job done.

Screenshot 2025-08-05 at 5 15 54 pm

oschwartz10612:
Exactly what I was thinking!

I know you have a couple PRs open right now. We will review this one and 
the others as soon as we can. We are swamped right now with stuff to do 
so I am sorry for the delays but we appreciate all of your work!

adrianeastles:
No rush @oschwartz10612, happy to bring features to this product that make the app more helpful.

oschwartz10612:
Thanks so much @adrianeastles appreciate it! 

We should probably update the new docs (docs-v2) with some obvious instructions about what to do - especially if there is a failure and the installer does not grab the token. People should be directed to go look at the logs. 

If you have some time to update that would be awesome otherwise we will do it ahead of the release.

oschwartz10612:
Also whoops merged a bit early. Could you leave a comment with the CLA please? Need that do be able to roll it into a release. Apologies! 

https://github.com/fosrl/pangolin/blob/main/CONTRIBUTING.md

adrianeastles:

oschwartz10612:
Thanks so much!

Issue #1184, opened 2025-07-31
Installer error if ipv6 is disabled


djlobster:
IPv6 is disabled on my VPS. Installer works fine if I say 'no' to starting the containers. If I say 'yes' then it gives me an error about the ipv6 portion because it's not available on my VPS. This also halts the installer so you never get asked about anything after this point like Crowdsec. I did originally comment out the IPv6 portion of the compose file and everything worked fine, I just realized after that it never did actually finish the installer. I tried it again just to be sure and then moved onto inputting 'no'.

```
=== Starting installation ===
Would you like to install and start the containers? (yes/no) (default: yes): yes
Pulling the container images...
[+] Pulling 3/3
 ✔ pangolin Pulled                                                                                                                                    0.1s 
 ✔ gerbil Pulled                                                                                                                                      0.2s 
 ✔ traefik Pulled                                                                                                                                     0.1s 
Starting containers...
[+] Running 1/1
 ✘ Network pangolin  Error                                                                                                                            0.0s 
failed to create network pangolin: Error response from daemon: Cannot read IPv6 setup for bridge br-9b656c609139: open /proc/sys/net/ipv6/conf/br-9b656c609139/disable_ipv6: no such file or directory
Error:  failed to start containers: exit status 1
```
After inputting 'no':
```
=== Starting installation ===
Would you like to install and start the containers? (yes/no) (default: yes): no

=== CrowdSec Install ===
Would you like to install CrowdSec? (yes/no) (default: no): yes
This installer constitutes a minimal viable CrowdSec deployment. CrowdSec will add extra complexity to your Pangolin installation and may not work to the best of its abilities out of the box. Users are expected to implement configuration adjustments on their own to achieve the best security posture. Consult the CrowdSec documentation for detailed configuration instructions.
Are you willing to manage CrowdSec? (yes/no) (default: no): yes
Stopping containers...
Installation complete!

To complete the initial setup, please visit:
https://domain/auth/initial-setup
```

Xentrice:
This is caused by my pull request #1111 
Damn, i was sure i tested everything.
I will create a pull request shortly making the IPv6 optional during installation. I didn't think this was necessary.
Which VPS and OS are you using?

djlobster:
[...]

PR #1212, opened 2025-08-04
make IPv6 optional during install

Xentrice:

## Description
Following #1111 and #1184.
Installer now asks whether the server is IPv6 capable. Does not disable IPv6 if user choses no, just doesn't enable the IPv6 network for docker. IPv6 connections will still succeed, but will get NATed to the Docker internal IPv4 address.
If needed, this can be changed to actually disable IPv6, for example by changing the Gerbil ports from 443:443 to 0.0.0.0:443:443 conditionally.

## How to test?
Test the install with both "yes" and "no".
"enable_ipv6 = true" should be included respectively not included in the docker-compose.yml

### DIFF
--- a/install/config/docker-compose.yml
+++ b/install/config/docker-compose.yml
@@ -60,4 +60,4 @@ networks:
   default:
     driver: bridge
     name: pangolin
-    enable_ipv6: true
+{{if .EnableIPv6}}    enable_ipv6: true{{end}}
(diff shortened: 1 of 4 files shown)


Xentrice:
Just noticed the conflict regarding file 4 - src/app/[orgId]/settings/sites/create/page.tsx
Can you resolve this, or should I create a new PR?

oschwartz10612:
Thanks so much!

Issue #1108, opened 2025-07-22
Feature Request: Move resource to another Org


nico-checkmk:
Hi,

I have a setup with several organisations, one for each domain. From time to time it comes to the case that I have to move a resource to another org. For now, I have to delete the entry and recreate it on the other side.
Is there a way to build a transfer option (maybe with a request what the new domain might be) into the GUI?

Greatings and props!

oschwartz10612:
Thanks for the suggestion! I think this is a good idea! I am going to mark this help wanted so if someone is interested in taking this we would be grateful! If not we will get to it at some point!

Pallavikumarimdb:
@oschwartz10612 , I want to work on this. Can you assign it to me?

oschwartz10612:
Thanks for the work!

Pallavikumarimdb:
> Thanks for the work!

Glad to contribute.

PR #1224, opened 2025-08-06
feat: Move resource to another org from Resource Info Box

Pallavikumarimdb:

## Description

- Enables resource move to a different org (via POST `/resource/:resourceId/move-org`)
- Displays other available orgs in ResourceInfoBox
- Core logic functional

Fix: #1108

### DIFF
--- a/server/setup/migrationsPg.ts
+++ b/server/setup/migrationsPg.ts
@@ -10,6 +10,7 @@ import m2 from "./scriptsPg/1.7.0";
 import m3 from "./scriptsPg/1.8.0";
 import m4 from "./scriptsPg/1.9.0";
 import m5 from "./scriptsPg/1.10.0";
+import m6 from "./scriptsPg/1.11.0";
 
 // THIS CANNOT IMPORT ANYTHING FROM THE SERVER
 // EXCEPT FOR THE DATABASE AND THE SCHEMA
@@ -21,6 +22,7 @@ const migrations = [
     { version: "1.8.0", run: m3 },
     { version: "1.9.0", run: m4 },
     { version: "1.10.0", run: m5 },
+    { version: "1.11.0", run: m6 }
     // Add new migrations here as they are created
 ] as {
     version: string;
--- a/server/setup/migrationsSqlite.ts
+++ b/server/setup/migrationsSqlite.ts
@@ -27,6 +27,7 @@ import m22 from "./scriptsSqlite/1.7.0";
 import m23 from "./scriptsSqlite/1.8.0";
 import m24 from "./scriptsSqlite/1.9.0";
 import m25 from "./scriptsSqlite/1.10.0";
+import m26 from "./scriptsSqlite/1.11.0";
 
 // THIS CANNOT IMPORT ANYTHING FROM THE SERVER
 // EXCEPT FOR THE DATABASE AND THE SCHEMA
@@ -53,6 +54,7 @@ const migrations = [
     { version: "1.8.0", run: m23 },
     { version: "1.9.0", run: m24 },
     { version: "1.10.0", run: m25 },
+    { version: "1.11.0", run: m26 }
     // Add new migrations here as they are created
 ] as const;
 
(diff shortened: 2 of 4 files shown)


oschwartz10612:
Looking good so far thanks for this pr! 

One thing to note: we are actively working on an update that moves the siteId from the resource to the target. So there is a  siteId on the target in the database and NOT on the resource. When you move a resource to a new org we might need to kick these sites off of the targets and show that somehow in the UI. We plan to have some of this merged from the `target-sites` branch soon.

Also one other thing I think you are thinking about a bit but is worth mentioning: when you move it the users and the roles and stuff assigned to that resource no longer apply because they are from the old org. So that all needs to be cleaned up in the database and I guess the new org will just have the current user with access. You can obviously keep the pin and pass and stuff. 

Let us know if you have any questions!

Pallavikumarimdb:
> Looking good so far thanks for this pr!
> 
> One thing to note: we are actively working on an update that moves the siteId from the resource to the target. So there is a siteId on the target in the database and NOT on the resource. When you move a resource to a new org we might need to kick these sites off of the targets and show that somehow in the UI. We plan to have some of this merged from the `target-sites` branch soon.
> 
> Also one other thing I think you are thinking about a bit but is worth mentioning: when you move it the users and the roles and stuff assigned to that resource no longer apply because they are from the old org. So that all needs to be cleaned up in the database and I guess the new org will just have the current user with access. You can obviously keep the pin and pass and stuff.
> 
> Let us know if you have any questions!

Thanks for the feedback!

I'll go ahead and implement the **Permission & Access Management part** — (Cleaning up the old org’s user/role associations from the resource and assigning the current user under the new org).  I'll make sure the pin and pass remain intact.

For the **Site-Target Relationship Handling**, I'll hold off until the updates from the `target-sites` branch are merged.

Pallavikumarimdb:
> Looking good so far thanks for this pr!
> 
> One thing to note: we are actively working on an update that moves the siteId from the resource to the target. So there is a siteId on the target in the database and NOT on the resource. When you move a resource to a new org we might need to kick these sites off of the targets and show that somehow in the UI. We plan to have some of this merged from the `target-sites` branch soon.
> 
> Also one other thing I think you are thinking about a bit but is worth mentioning: when you move it the users and the roles and stuff assigned to that resource no longer apply because they are from the old org. So that all needs to be cleaned up in the database and I guess the new org will just have the current user with access. You can obviously keep the pin and pass and stuff.
> 
> Let us know if you have any questions!

Hi @oschwartz10612 ,
Quick question: when moving resources between orgs, there may be cases where we need to “kick off” sites from targets (so the target remains but without a site attached). For that to work, I think we’ll need to make siteId nullable in the target schema.
Schema change (making siteId nullable) is required if we want to preserve targets instead of deleting them — otherwise we’re forced to drop the target entirely when the site is removed from moving resources.

Do you agree that updating the schema this way is the right approach, or do you see a different direction? Just want to confirm before making changes. Would love your thoughts.

oschwartz10612:
Hi sorry for the delay!

Yeah I agree I think we could do a schema change to make it null able. Its probably better it does not get kicked off. 

There may be some knock on effects of making this null able but I will do my best to handle those after merging.

oschwartz10612:
Few more things. 

I think we should move the picker into the general settings. I think this would be more consistent with how things work elsewhere.

image

Also I think when it is successful this should probably be a toast and not an alert - I dont think we use the alert anywhere else. 

image

Sorry for the change requests. Thanks for taking a stab at this!

Pallavikumarimdb:
Hi @oschwartz10612 , I’ve made the requested change. Could you please review and let me know if any further adjustments are needed?

oschwartz10612:
@Pallavikumarimdb Thanks for the work on this one! Sorry not taking. We think this will be too complicated to maintain as the resource continues to evolve and more things get connected to it like polices, roles, etc... It seems it was already quiet difficult to calculate this so we dont want that to cascade. There were also concerns around domains and stuff moving cross org.

We think the recommended procedure for now would be to delete from one org and then recreate in the next.

PR #1246, opened 2025-08-10
Update express to version 5

Lokowitz:

## Description
Fixes the issue with PR https://github.com/fosrl/pangolin/pull/944/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

See [documentstion](https://expressjs.com/en/guide/migrating-5.html#path-syntax)

please also notice and adapt if needed:
> *splat matches any path without the root path. If you need to match the root path as well /, you can use /{*splat}, wrapping the wildcard in braces.


oschwartz10612:
Thanks so much I appreciate you doing the research for that!

PR #1231, opened 2025-08-07
Add rule templates for reusable access control rules 🚀

adrianeastles:
### Description

This PR adds a new rule templates feature that allows users to create reusable sets of access control rules that can be assigned to multiple resources. This provides a more efficient way to manage consistent access control across multiple resources.

### TODO
1. Notify users when templates are updated
2. Add more granular loading states for template operations
3. Improve error handling and recovery for template operations
4. Add confirmation dialogs for destructive actions
5. Improve success notifications and feedback
6. Implement pagination for template rules table
7. Implement pagination for resource rules table

### Key Features

- **Rule Templates Management**: Create, edit, and delete rule templates
- **Template Assignment**: Assign templates to resources, automatically creating resource rules from template rules
- **Automatic Propagation**: Changes to template rules automatically propagate to all assigned resources
- **Template Rule Protection**: Rules created from templates are protected from editing (except priority) on resource pages
- **Visual Indicators**: Clear badges distinguish between template-managed and manual rules

### Technical Implementation

- New database tables: `ruleTemplates`, `templateRules`, `resourceTemplates`
- Added `templateRuleId` field to `resourceRules` for tracking template origins

### Database Changes

- Migration script `1.10.0` adds new tables and `templateRuleId` column

### 1. Create a Rule Template
1. Navigate to `http://localhost:3002/org/settings/rule-templates`
2. Click "Create Template" and fill in name/description
3. Verify template appears in the list

### 2. Add Rules to Template
1. Click on a template to edit it
2. Go to the "Rules" tab (default)
3. Add rules with different match types (IP, IP Range, Path)
4. Verify rules appear in the table and can be edited/deleted

### 3. Assign Template to Resource
1. Go to a resource's rules page: `http://localhost:3002/org/settings/resources/{resourceId}/rules`
2. Enable rules if not already enabled
3. In the "Rule templates" section, assign a template
4. Verify template rules appear in the resource rules table with "Template" badges

### 4. Test Template Rule Protection
1. On the resource rules page, try to edit a template-managed rule
2. Verify only priority can be changed, other fields are disabled
3. Verify delete button is disabled for template rules

### 5. Test Template Rule Propagation
1. Edit a rule in the template
2. Verify the change propagates to all assigned resources
3. Verify the resource rules show the updated values

### 6. Test Template Unassignment
1. Unassign a template from a resource
2. Verify all template-managed rules are removed from the resource
3. Verify manual rules remain unchanged

#### Rule Template Screens
[...]

PR #1248, opened 2025-08-10
Enhancement #906/Add Shared IP Rule Sets, Auto CIDR Handling, and Preserve Match Type

### LINKED ISSUE #906: [Feature Request] feedback after migration: ip rules set, parsing of ips, autosaves and more
# Feedback After Migrating from NPM to Pangolin

# Migration Context

I migrated from NPM primarily due to the lack of timely follow-up on CVEs. In contrast, I’ve taken a strong liking to this project, which I view as a modern and cleaner GUI frontend for Traefik.

I’m not personally interested in the built-in VPN/tunnel functionality  I prefer to manage my own networking and tunnels but I did still install the full VPN stack for potential testing down the line. The SSO login feature, however, is a welcome addition to my stack.

---

## Setup Notes

I started from scratch using only Pangolin and its documentation nothing else i avoided the recommended youtube videos.  
I chose the manual setup path because I avoid automated Docker install scripts I like to know *exactly* where files go and how everything connects.

**Specificity:**

I assigned a dedicated IP to the reverse proxy container. This eliminates the need to expose ports at the Docker level. With this setup, I route all internal TCP/UDP traffic via DNS wildcards to the reverse proxy, which then forwards it to the correct internal hosts. External/public traffic continues to be routed through my main edge router.

In my opinion, this should be the default recommended installation method, as Pangolin functions more like a virtual appliance than a traditional ephemeral Docker container (I can already hear the security engineers screaming in the background).  
This setup would also solve the current issue of needing to expose ports on the Docker level every time a new raw resource is added.
But that's just my grain of salt not a formal request. 🧂

```yaml
gerbil:
  image: fosrl/gerbil:1.0.0
  container_name: gerbil
  restart: unless-stopped
  networks:
    default: {}
    br0:
      ipv4_address: 192.168.1.6
```

**What i have lost:**

- No more static webpage hosting, while i understand that Traefik does not support hosting of HTML files this is still something i have now to spin up a dedicated infrastructure for.
- No more single cert for both *.basedomain.com & basedomain.com this is annoying if i have another service that needs this i need another cert generation.

For example to replace the static webpage hosting:

```yaml
  cv:
    container_name: nginx
    image: nginx:alpine
    ports:
      - 8087:80
    volumes:
      - ${path_root}nginx/:/usr/share/nginx/html:ro
```

---

## Here’s a the big list of Improvements That Would've Made My Migration and Usability Smoother:

---

### 1.1 Shared IP Rule Sets  
Most of my hosts use common rules (e.g., allow local IPs, office IPs, VPN).  
Currently, rules are set **per-host only**, but **global/shared IP rules** would save a ton of time.  
- NPM handles this via reusable IP lists.
- Please add the ability to create and reuse **IP allow/deny lists**.

IP Rules

Missing IP Lists

### 1.2 IP Range vs. Single IP  
Don’t make the user choose between "IP" and "IP Range".  
If someone enters `192.168.1.2`, just assume `/32` automatically.  

### 1.3 Add rule changes match type
Pressing the Add rule button changes the Match type this is annoying when adding big quantities of rules

Image

---

### 2.1 Smart Parsing for Host Targets  
When pasting a URL like `http://192.168.1.2:81`, auto-split it into:
- Protocol: `http://`
- Host/IP: `192.168.1.2`
- Port: `81`

…and strip out the extras.  

URL Input

### 2.2 Multiple Hostnames for a Single Resource  
Let me define multiple domains (`example.com`, `www.example.com`) on the **same resource**.  
Duplicating config for every alias is a chore.

Multiple Hosts

---

### 3.1 Tab Switching = Lost Changes  
When switching tabs on a resource config, changes get lost if you forget to hit "Save" first. That’s painful.

### 3.2 Leave Warning  
Add a `beforeunload` browser prompt when unsaved changes exist:  
> “Are you sure you want to leave this page?”  

[MDN Docs for reference](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event)

### 3.3 Merge Tabs into One Page  
Instead of hiding settings across the 4 tabs "General Proxy Authentication Rules", just give me a readable single-page layout.  
Clicking back and forth is inefficient, and vertical scroll is not a sin even tho there are plenty of dead spaces in these tabs.

### 3.4 Too many clicks for a new hosts

Same-ich as 3.3 but different in the ui currently so it is another point. 

When deploying a new resource you first have to save the title and host before being granted access to the settings thats not necessary and all options should be there from the get-go.

---

### 4.1 Show Target in the Dashboard  
Add a "Target" column to the main Resources page so I can quickly see what’s pointing where.

Dashboard Targets

### 4.2 Customizable Columns  
Even better: let users toggle which columns are visible.

### 4.3 Status icons  
Next to each resource, display a small status indicator if the target is not responding to a simple TCP check.  
Not asking for full-blown monitoring  just a quick QOL check that's refreshed when the UI reloads.  
It would make spotting downed hosts much easier at a glance.

Image

### 4.4 Does not save amount of entries

This value is never saved on any of the dashboard... do you want me to pull my eyeballs out ? i have to set it after each refresh of a page this is beyond sadistic UX design.

Image

### 4.5 Weird "Enter" behaviors

I have on almost all fields tried to type press the `Enter` key and almost every time the result was... weird.
For example in this case pressing `Enter` refresh the page in place of submitting the entry.

Image

---

## 5. Docker Label-Based Config

Why not support full config via Docker labels (e.g. like Traefik)?  
This would simplify automation in container-heavy setups. In my case i would have used this for all my containers and the ui for non docker hosts.

---

## 6. SSL Management from the GUI

Managing multiple SSL certificates from multiple config files is a pain in the bottom.  
There should be a centralized SSL management UI in the panel.

For exemple:
Image

If Traefik’s certificate handling is the blocker maybe consider ditching Traefik in favor of a more flexible solution? (it is a supposition i do not know)

Image

Image

---

## 7. Raw TCP/UDP

Adding automatically the entries in traefik_config.yml for all the resource would be nice.

---

## 8. Health check

Gerbil does not have an included health check maybe add this to the default compose and traefik too ?

Image

---

## 9. Path for Traefik add-ons ?

While the project support Crowdsec there is no documentation on how to implement it if you followed the manual setup path.
This is the documentation:
``
Installation.
Crowdsec can be installed using the Pangolin Installer.
``

---

## 10. Customizable SSO page

The SSO login page should be customizable to better match local branding.

Ideally, this could be done in one of two ways:
- Simple Web UI (like UniFi): change logo, background, text, colors.
- HTML/CSS Template in the config directory: for full control over layout and styling.

Either option would make the login experience feel more homey, especially in internal or multi-user environments.
My personal preference would be the full control over the html page. 

Image

Image

---

## 11. Simple redirections

A simple UI to create resources that would forward one host to another, similar to what NPM proposes.

Image

---

## 12. Custom HTML template for errors

Who doesn’t like to have custom error pages? It’s fun, and a framework to implement it would be NOICE.  
The default 404 is making me want to help it end its days.

All of this would allow you to have by default pangolin branded error pages 🤫 And get a famous artist to draw-up your default errors
Pretty sure some would kill to get commissioned for that. 
here a small idea of what could be done:

Image 

Image

inspiration from:
https://happydorid.tumblr.com/image/106988062584

Here is the traefik documentation for this: 
https://doc.traefik.io/traefik/middlewares/http/errorpages/

This is what I had done on nginx:
```nginx
root /websites;
location / {
    error_page 400 /error_pages/HTTP400.html;
    error_page 401 /error_pages/HTTP402.html;
    error_page 402 /error_pages/HTTP402.html;
    error_page 403 /error_pages/HTTP403.html;
    error_page 404 /error_pages/HTTP404.html;
    error_page 500 /error_pages/HTTP500.html;
    error_page 501 /error_pages/HTTP501.html;
    error_page 502 /error_pages/HTTP502.html;
    error_page 503 /error_pages/HTTP503.html;
    proxy_intercept_errors on;
}

location /error_pages/ {
    alias /websites/error_pages/;
    internal;
}
```
## 12. RR Random Rambling

Nowhere in the docs is it mentioned that the Traefik dashboard is enabled on port 8080.  
This should either be clearly documented... or not exposed at all by default.

If it's meant to stay, it should ideally be served at something like `https://pangolin.example.com/traefik-dashboard` and integrated into the admin UI with a quick-access button that would iframe it onto the ui.

Image

Here’s how I’ve personally handled it:

Image

---

Maybe out of scope, but worth thinking about:

- Auto-scan `docker.sock` for containers not routed through Pangolin, then show them in a UI to quickly pick a port and auto-fill a resource. One click = done.
- Or just offer prebuilt labels that can be added to a `docker-compose` file to declare resources inline.

Either would then make adding services way faster and brain-dead simple. 

---

Why not either integrating or building an alternative to Anubis into the sso page ?

https://anubis.techaro.lol/

also need a solution for this issue:

https://anubis.techaro.lol/docs/admin/configuration/custom-status-codes/

---

unified robots.txt, humans.txt, security.txt, etc etc
it would be nice to configure these in pangolin and then assign them to each ressource.

---

was playing around with OICD and thought this would be nice to have people request access that you could acknowledge manually 
Image

---

## Final Word

**Good job on the software.** I'm happy with the result and hope this feedback is taken constructively.  
I’m optimistic about the future of the project and will probably buy a supporter key once my setup is fully stable.
👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍
😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘😘

Pallavikumarimdb:

## 1. IP Rules: Global Sets + Usability (Fixes #906 )

1.1 **Shared IP Rule Sets (Global Lists)**

   * Allows users to create reusable IP allow/deny lists (`IP_SET`) that can be applied to multiple resources.
   * Backend stores only the `ipSetId`; actual matching is resolved via the IP set contents.
   * UI shows IPs from the selected set for clarity, but backend still uses `ipSetId` for matching.

1.2 **Automatic CIDR for Single IPs**

   * No need to choose between `"IP"` and `"CIDR"` manually.
   * Single IPs (v4/v6) are auto-normalized to `/32` or `/128`.
   * CIDRs are accepted as-is.

1.3 **Preserve Match Type when Adding Rules**

   * When clicking “Add rule”, the current match type remains unchanged.
   * Prevents accidental resets when entering multiple rules.
  
image

image

image

image

This is a sub-issue of #906


Pallavikumarimdb:
@oschwartz10612 , Please take a look and let me know if any changes or improvements are needed. I’ll be creating a separate PR for each sub-issue from #906

adrianeastles:
@Pallavikumarimdb does your first feature overlap with this current PR #1231 ?

Pallavikumarimdb:
Hi @adrianeastles , Thanks for the context! My feature focuses on reusable IP sets, which can be used within rules. #1231  rule templates feature manages reusable sets of rules, so I see them as complementary. IP sets from my first feature can be leveraged inside rule templates to make rule management more efficient. Happy to collaborate!

adrianeastles:
Just putting my two cents out there from a UX/Software engineer background.

Both PRs are essentially solving the same user problem- helping users manage and reuse rules more efficiently - but they're taking different approaches:

PR #1248 (IP Sets) creates reusable collections of IP addresses that can be referenced in rules:
- "Office IPs" = `[192.168.1.0/24, 10.0.0.0/8]`
- Rules can then reference this IP Set instead of duplicating IPs

PR #1231 (Rule Templates) creates reusable collections of entire rules:
- "Security Rules" = `[ACCEPT office IPs, ACCEPT VPN ranges, DROP admin paths]`
- Resources can apply the whole template at once

Here's where it gets interesting - Rule Templates can already do what IP Sets do! You could create a template with multiple IP-based rules that achieves the same reusability as an IP Set. So we need to ask:

- Do we need both features, or is one redundant.
- Would users prefer IP Sets or just Rule Templates?

Pallavikumarimdb:
Hi @oschwartz10612 , We’re discussing the overlap between the IP Sets feature and the Rule Templates PR. Both aim to improve rule reusability but at different levels - IP Sets manage reusable IP groups, while Rule Templates handle full rule sets. Do you have any thoughts on whether both should coexist or if one approach is preferred for the project’s direction?

P6g9YHK6:
I would be happy with either approach but got a slight preference for Rule Templates as it would be easy to have displayed on the dashboard what template is assigned to who.

oschwartz10612:
@Pallavikumarimdb as usual sorry for all your hard work on this then to close! We are planning on doing resource policies which maps better to #1231 so I think we are going to take that. Managing sets of IPs as they need to get blocked would be covered in creating a resource policy with a bunch of rules and applying it.

curious-debug:
Hey @oschwartz10612, in your plan to create a resource policy system (which is a good idea by the way), please incorporate this: 
- allow admin to choose what happens when visiting a resource fails all IP Rules at the outset (that is, prior to Authentication). I received some feedback and included an "Updated Recommended Action" section in [the original post](https://github.com/fosrl/pangolin/issues/1614). (also copied below)
-----
Per insightful discussion ... the best path forward is this:

When not satisfying any IP Rules, admin should be able to select one of these for what happens next:

1. send client to Auth page;
2. select from list of HTTP status codes to send to client (e.g.: 401 Unauthorized, 403 Forbidden, 404 Not Found, 405 Method Not Allowed, 503 Service Unavailable, custom status code, etc.); or
3. drop connection (send no response to client).

My original post primarily suggested sending no response, so as to not inform the failed IP visitor that the resource exists. @hhftechnology suggested keeping the 401 for its use case. Thus, the best option is to allow the admin flexibility to choose the course of action when visiting a resource fails all IP Rules, allowing the admin to pursue further logging/tracking and/or security, as best determined by the admin.

PR #1281, opened 2025-08-14
fix: change default integration_api to 3004

jackrosenberg:

## Description

Currently both gerbil and the integration api server listen on port 3003. This causes site connectivity to break, since gerbil is not reachable. I would propose changing the API server default port to 3004.
closes https://github.com/fosrl/gerbil/issues/18

### DIFF
--- a/server/lib/readConfigFile.ts
+++ b/server/lib/readConfigFile.ts
@@ -43,7 +43,7 @@ export const configSchema = z
         server: z.object({
             integration_port: portSchema
                 .optional()
-                .default(3003)
+                .default(3004)
                 .transform(stoi)
                 .pipe(portSchema.optional()),
             external_port: portSchema


oschwartz10612:
Only concern with this is it might break people's integration APIs if they do not define this in the config file. But I agree I do see the issue here...

Maybe we make it clear in the release notes?

oschwartz10612:
I suppose we could try to migrate people's config but that could get messy

jackrosenberg:
> I suppose we could try to migrate people's config but that could get messy

Is that something we can easily do? If not, I think it would probably be best to just rip off the bandaid. I can PR the relevant docs tonight/tomorrow.

jackrosenberg:
>I can PR the relevant docs tonight/tomorrow.

https://github.com/fosrl/docs-v2/pull/4

v1rusnl:
Why migrate people configs? The API is disabled by default as far as I can see? I assume people who activate it know what they do. @jackrosenberg changes to the docs + a prominent notice in one of the next release notes explaining why in a short term should do it imo.

oschwartz10612:
Agree sounds like a plan! I will highlight in the release notes!

PottiMc:
@oschwartz10612 
The documentation needs to be adjusted according to this change: https://docs.digpangolin.com/self-host/advanced/integration-api

jackrosenberg:
> @oschwartz10612 The documentation needs to be adjusted according to this change: https://docs.digpangolin.com/self-host/advanced/integration-api

fixed this in: https://github.com/fosrl/docs-v2/pull/4

v1rusnl:
@oschwartz10612 
Do we still need 
`server:
  integration_port: 3004`
in config.yml or just 
`flags:
  enable_integration_api: true`

Looking at the latest config template, it seems to be much much smaller than my inital config from 1.2.0 when I installed it. I deleted a lot of stuff in it testwise yesterday and it seems Pangolin works as before even config.yml is now half the size than before. I also deleted the integration port directive and API works nonetheless.

jackrosenberg:
> Do we still need `server: integration_port: 3004` in config.yml

Nope, but it's nice in the docs to let people know it's an option. Also yes, most of the config is superfluous. For a true minimal config just delete it all, and pangolin will yell at you telling you exactly what fields are missing :)

v1rusnl:
> > Do we still need `server: integration_port: 3004` in config.yml
> 
> Nope, but it's nice in the docs to let people know it's an option. Also yes, most of the config is superfluous. For a true minimal config just delete it all, and pangolin will yell at you telling you exactly what fields are missing :)

Just made a discussion since I recognized the PR is closed while you answered 😆 
https://github.com/orgs/fosrl/discussions/1338

But thanks for clarification! Perhaps the discussion is helfpul for others in a more prominent way.

PR #1283, opened 2025-08-15
Enhancement #906/UX & Usability Improvements - Prevent Data Loss, Persist Preferences

Pallavikumarimdb:

## Description

This PR addresses several usability issues highlighted in #906  to improve user experience and prevent accidental data loss. It focuses on UX reliability and reducing user frustration.

**Issues & Sub-issues addressed:**

* **3.1 — Tab Switching = Lost Changes**
  Switching between tabs on a resource configuration currently causes unsaved changes to be lost if the user hasn’t clicked “Save” first.

* **3.2 — Leave Warning**
  No warning is shown when navigating away or closing the browser with unsaved changes.

* **4.4 — Does Not Save Amount of Entries**
  The “entries per page” setting is not remembered and resets after every page refresh.


oschwartz10612:
Milo and I reviewed this and we are worried that it will be too much overhead throughout the application to worry about unique implementations everywhere to check on the tab switching and lost changes issues. We are leaning toward at least for now just keeping it as up to the user to save before moving and to know when to save. We are thinking some of the forms will be redone in the future to make this more clear.

So I hate to ask you to throw away work but I dont think we are going to take 3.1 and 3.2 right now.

But for 4.4 that's an easy one - we do want to keep that! Are you maybe able to break it out into a seperate PR or change this PR to just have the fix for that?

Pallavikumarimdb:
Thanks for the feedback! I completely understand the concern around 3.1 and 3.2 and the potential overhead.

I've extracted 4.4 into a separate PR: #1395 .
That PR is now focused solely on persisting the "entries per page" preference.

oschwartz10612:
I think we are going to close this one for the above reasons

PR #1316, opened 2025-08-22
Enhancement #906/Smart Host Parsing for Target Address

Pallavikumarimdb:

## Description
This PR introduces smart host parsing for the Target Address field (issues highlighted in #906 ).
**Sub-issues addressed: 2.1 Smart Parsing for Host Targets**

**Solution:**
Users can now paste a full URL (e.g., http://192.168.1.2:81) directly into the IP/Hostname input, and the system will automatically extract and populate method, host, and port fields. Users still have the flexibility to manually edit each field if required.

* Added a parseHostTarget utility to extract protocol, host, and port from a pasted/typed URL or raw IP input.
* Auto-assign default ports when user input does not include one:
   - http:// → port 80
   - https:// → port 443
* Preserves user-provided ports if specified.
* Works for both domain names and IP addresses.

### DIFF
--- a/server/lib/remoteProxy.ts
+++ b/server/lib/remoteProxy.ts
@@ -70,4 +70,4 @@ export const proxyToRemote = async (
             )
         );
     }
-}
+};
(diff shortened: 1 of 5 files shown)


oschwartz10612:
Thanks so much! Seems to work well! :}

PR #1328, opened 2025-08-24
Enhancement #906/Resources Dashboard: Targets Column, Customizable Columns & Status Indicators

Pallavikumarimdb:

## Description
This PR implements the following enhancements to the Resources page, addressing issues #906 .

**4.1 Show Target in the Dashboard**
- Added a new “Target” column to the Resources table.
- Displays all configured targets for each resource in a dropdown menu.
- Targets are copyable (ip:port) with a single click.
image

 **4.2 Customizable Columns**
- Introduced column toggle controls, allowing users to select which columns are visible.
- Improves usability for large orgs with many resources.
image
image

**4.3 Status Icons**
- Added online/offline indicators (green/red dots):
> - Per resource: aggregated status → "online" if at least one target is reachable.
> - Per target: individual status displayed next to each target in the dropdown.
- Status is determined via a lightweight TCP connectivity check (tcp-check-batch endpoint).
- Results refresh on page reload; not intended as full monitoring, just quick QOL visibility.
image

### DIFF
--- a/src/components/DataTablePagination.tsx
+++ b/src/components/DataTablePagination.tsx
@@ -24,7 +24,6 @@ interface DataTablePaginationProps<TData> {
     isServerPagination?: boolean;
     isLoading?: boolean;
     disabled?: boolean;
-    renderAdditionalControls?: () => React.ReactNode;
 }
 
 export function DataTablePagination<TData>({
@@ -34,8 +33,7 @@ export function DataTablePagination<TData>({
     totalCount,
     isServerPagination = false,
     isLoading = false,
-    disabled = false,
-    renderAdditionalControls
+    disabled = false
 }: DataTablePaginationProps<TData>) {
     const t = useTranslations();
 
@@ -115,11 +113,6 @@ export function DataTablePagination<TData>({
                         ))}
                     </SelectContent>
                 </Select>
-                {renderAdditionalControls && (
-                    <div className="flex items-center space-x-2">
-                        {renderAdditionalControls()}
-                    </div>
-                )}
             </div>
 
             <div className="flex items-center space-x-3 lg:space-x-8">
--- a/src/components/ui/data-table.tsx
+++ b/src/components/ui/data-table.tsx
@@ -9,9 +9,7 @@ import {
     SortingState,
     getSortedRowModel,
     ColumnFiltersState,
-    getFilteredRowModel,
-    VisibilityState,
-    Column
+    getFilteredRowModel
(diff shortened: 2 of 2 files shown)


oschwartz10612:
This is great! I think we want to take this one but I think we are not ready yet for the status reporting and checking and stuff - and this could conflict with the health checks we recently added to the cloud. Do you think you could _remove_ the status online/offline column and the status indication for each of the targets? If so that would be great and then I think we could merge! I am sorry to ask you to remove work.

I also think if we are going to do column filtering we should do it on all of the tables instead of just the resource table. And maybe we should move this button into the footer so it is not front in center for the user at the top? I think most of the time people will not want to change the columns but I think it is important that there is the ability.

Pallavikumarimdb:
@oschwartz10612 , I’ve made the change. Could you please review and let me know if any further adjustments are needed?

P6g9YHK6:
> This is great! I think we want to take this one but I think we are not ready yet for the status reporting and checking and stuff - and this could conflict with the health checks we recently added to the cloud. Do you think you could _remove_ the status online/offline column and the status indication for each of the targets? If so that would be great and then I think we could merge! I am sorry to ask you to remove work.
> 
> I also think if we are going to do column filtering we should do it on all of the tables instead of just the resource table. And maybe we should move this button into the footer so it is not front in center for the user at the top? I think most of the time people will not want to change the columns but I think it is important that there is the ability.

i do not think a status indicator would conflict with the cloud feature as you are just displaying if the backend is up or not.
if not your failovers or any cloud feature would not work either.

i also feels like gate-keeping when the code is there and functional,  people using self-hosted do not deserve to have QOL feature kept away from them because it is somewhat similar to another service you sell. 😤

oschwartz10612:
Hey @Pallavikumarimdb! With the health checks out in oss we think we should complete this PR! Some notes we think we would like to change:

- [ ] Combine the targets into the status column. Maybe when you hover/click on the status icon it shows the targets? 
- [ ] Use the health check status for the status of the target in the display
- [ ] Targets can have unknown/not configured health checks. We should show these as gray icons
- [ ] We should summarize the status icon with the info from the targets. So if all of the targets are green then the status is green, if some are red then it is a "degraded status" and the color is yellow or something with a different icon. If they are all red then the status should indicate that its fully offline. In the calculation maybe we omit the unknowns/no monitored targets.

oschwartz10612:
Thanks so much! :}

Issue #1369, opened 2025-08-28
Can't create subdomain resource


nico-checkmk:
Hi again,

I have a newly installed single instance (fresh with 1.9.0) on which I try to create resources for a subdomain.
I have added my domain and it's subdomain to the configuration (example.com and sub.example.com). Then I create a resource for name.sub.example.com, click on create and the URL is only set to sub.example.com. Also traefik is trying to create a certificate for sub.example.com
Until version 1.9.1, it was possible to manually set the correct domain when editing the resource, but since the update to 1.9.2, it is not possible anymore.
When I edit my resource, go to the "General" tab and edit the Domain (which shows https://sub.example.com), it falls instantly back to sub.example.com (even without https)

Pallavikumarimdb:
Hi @nico-checkmk , What you’re experiencing is related to issue #1373 . Right now, if you enter the subdomain first and then choose the base domain, only the base domain gets saved - which is why you’re seeing it fall back to just sub.example.com. The behavior works correctly if the base domain is selected first and then the subdomain is entered.

nico-checkmk:
Thanks a lot!
My workaround for now is to enter the hostname and the subdomain part into the text field and then select my base domain. This looks a bit strange, but for now it works.

<img width="1541" height="193" alt="Image" src="https://github.com/user-attachments/assets/9668125a-27f2-4ddf-87f5-610e3aa762f5" />

oschwartz10612:
Thanks to @Pallavikumarimdb will be fixed in next patch release! :}

PR #1375, opened 2025-08-28
Preserve subdomain input when selecting a base domain

### LINKED ISSUE #1373: Subdomain not saved if entered before selecting base domain when editing a resource
When editing a resource domain, the subdomain is not saved correctly if the user enters it before selecting the base domain. If the base domain is selected first, and the subdomain is entered afterward, it works correctly.

Pallavikumarimdb:

## Description
This PR resolves #1373 and #1369 , where a subdomain entered before selecting a base domain was not saved when creating or editing a resource.

**Changes introduced:**

- Users can safely type a subdomain before choosing a base domain without losing input.
- Subdomain change and base domain selection now both trigger validation → ensuring state stays in sync regardless of order of operations.
- Improved subdomain validation to ensure consistency across inputs.
- The subdomain is sanitized (invalid characters removed).
- If compatible, it is preserved and combined with the base domain.

### DIFF
--- a/server/lib/schemas.ts
+++ b/server/lib/schemas.ts
@@ -1,18 +1,24 @@
 import { z } from "zod";
 
+
 export const subdomainSchema = z
     .string()
     .regex(
-        /^(?!:\/\/)([a-zA-Z0-9-_]+\.)*[a-zA-Z0-9-_]+$/,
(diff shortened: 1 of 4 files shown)


oschwartz10612:
Thanks so much Pallavi!

PR #1383, opened 2025-08-29
Fix subdomain preservation in DomainPicker when switching base domains

arevindh:

## Description
When a user typed a subdomain and then selected a different base domain, the subdomain value was being cleared due to `onDomainChange` being called with `subdomain: undefined`.

## How to test?

- User need to to have more than 1 domain to test.
- Type subdomain first then change the base domain.
- If the subdomain is valid for the new base domain, the subdomain should be preserved

This should be tested on **Create Resource Page** and **General Section** of  **Resource Settings**


Pallavikumarimdb:
Hi @arevindh , I had a look at your fix. I see you’re clearing the subdomain input whenever the base domain is edited after a subdomain has already been entered. That definitely avoids the case where only the base domain gets saved, but it puts the burden back on the user since they need to re-enter the subdomain.

I’ve already opened a PR with an alternative approach here: #1375 for similar issue. Could you verify if this resolves your concern as well?

arevindh:
> Hi @arevindh , I had a look at your fix. I see you’re clearing the subdomain input whenever the base domain is edited after a subdomain has already been entered. That definitely avoids the case where only the base domain gets saved, but it puts the burden back on the user since they need to re-enter the subdomain.
> 
> I’ve already opened a PR with an alternative approach here: #1375 for similar issue. Could you verify if this resolves your concern as well?

I haven't looked at your PR, but the subdomain does not get cleared if its valid.  

pangolin-1

oschwartz10612:
Thanks so much @arevindh but I think it was solved in [#1375](https://github.com/fosrl/pangolin/pull/1375). I appreciate the contribution!

Issue #1393, opened 2025-09-01
When clicking on the IP/Hostname field and saving, the Port and Method gets reset to default.


FabianBartl:
I updated to Pangolin v1.9.3 today, and wanted to change an IP/Hostname field of a Resource, but after saving the changes I got a Bad Request error.

The Issue is, that the Method and Port fields are reset to their default just by clicking in the IP/Hostname field and then somewhere else outside it, like saving the changes.

Pallavikumarimdb:
After recent change where the system auto-parses full URLs from the IP/Hostname field. It looks like the parse host target logic is being triggered even when values for method/port are already present, which causes them to reset to defaults unexpectedly.

I’m looking into this and will fix soon so that existing values for Method and Port are preserved unless the user explicitly changes them.

PR #1394, opened 2025-09-01
Fix/hostname field reset port and method

Pallavikumarimdb:

## Description
Modified the handlers to only trigger URL parsing when the input actually looks like a URL.

**Conditional updates:**
- Method: Only update if protocol was explicitly provided OR no method currently set
- Port: Only update if port was explicitly provided OR no port currently set
- IP: Always update (this is the primary field)

**Preserve existing values: Plain IP/hostname entries don't trigger parsing at all.
Update the parseHostTarget to support h2c.**

Fixes: #1393


oschwartz10612:
Awesome! Thanks for following and fixing so quickly! I tested and confirmed it works! Will release a patch later.

Pallavikumarimdb:
Happy to contribute! thanks for the quick review and support.

Issue #1293, opened 2025-08-17
Maxing out resources to whatever the user has (or 100). Making resources browsable under sites?


MrKromz:
I love pangolin!!!

Hello, curious what other people thought.... first time on github so apologies for not being so technical in advance.
I run an instance of newt for every server I use. I have a lot of resources (40+ and growing). 

1. is there a chance you folks could make the resources viewable under each site?  I spend most of my time on pangolin trying to find my resource. Having things like my truenas or unraid servers get pushed to the bottom of the list by default (alphabetical)

2. is there a chance you could make the resource page open to max number of resources (not just 10). I keep finding myself scrolling to the bottom and opening 100 as the max resources just to find a resource.

I am thinking something like a directory tree....
1.domain (optional as some people have multiple domains)
2.site
3.resources

I hope this makes sense... I dont even know if others have this same experience or not.....

oschwartz10612:
Hi!

Thanks for the feature suggestion!

I definitely see the need for this especially with that many sites and 
resources.

FYI - in version 1.9.0 we are removing sites from resources and adding 
them to targets so you can have different targets on different sites for 
the same resource.

PR #1421, opened 2025-09-05
Enhancement/Making resources browsable under sites

Pallavikumarimdb:

## Description
**Implementing site-specific resource browsing.**

- Site-specific resource listing: Shows only resources that target the selected site
- Search and filtering: Users can search through resources for that site

Fixes: #1293


Pallavikumarimdb:
Hi, @oschwartz10612 , Please take a look and let me know if any changes or improvements are needed.

oschwartz10612:
Thanks for the work on this @Pallavikumarimdb! We think it has drifted a little bit too much with all of the new information and it requires us to make sure we are putting resource info in two places now. We also are thinking about a future graphic implementation where you can see everything in one place separate from the table views and we feel it would be better to tackle it then.

PR #1430, opened 2025-09-07
Fix: Include "region" field in ExitNode query to match schema

Pallavikumarimdb:

## Description
This PR fixes a type mismatch issue in `listExitNodes`. The `ExitNode` type expects a `region: string | null` field, but the query did not select it, causing errors.


oschwartz10612:
Thanks! :} 

Shouldn't be pushing small changes to dev like this without testing first

Issue #1426, opened 2025-09-07
Validation error of IP address on Client Resource: "Invalid IP at destinationIp"


billyinferno:
When you create Client Resource and put "localhost" on the Destination IP, it will works.

But if you edit the same Resource and put "localhost" it will showed Validation Error.

Pallavikumarimdb:
This looks similar to #1419

oschwartz10612:
Yeah thanks @Pallavikumarimdb

Issue #1419, opened 2025-09-05
Edit client resource from IP field does not accept `host.docker.internal` (but creating does)


JobDoesburg:
The title pretty much explains it.
When editing a client resource, the IP address field does only accept real IP addresses.
When _creating_ a new client resource, a value such as `host.docker.internal` _is_ accepted (which should be the expected behavior).

oschwartz10612:
We have completely redone the client resources. This will be fixed in 1.13.0 mostly by it functioning differently

PR #1431, opened 2025-09-07
Fix inconsistent destinationIp validation between create and update APIs

Pallavikumarimdb:

## Description
This PR resolves the inconsistency reported in #1419 and #1426 where the update site resource rejected hostnames (localhost, host.docker.internal, etc.) due to strict .ip() validation, while the create API accepted them.


oschwartz10612:
Thanks so much Pallavi!

PR #1488, opened 2025-09-18
Update de-DE.json

cku-heise:

## Description
"Spielpfad" is literal, but wrong translation (not a word in German). "Unterverzeichnis" would be the best approximation of the UI label here. The translation omits the word "match" ("Prüfen" or something) for the sake of brevity, but combined with the "+" should be informative enough.


oschwartz10612:
Thanks!

PR #1550, opened 2025-09-27
Add minor version tags to Docker build commands in Makefile

Shamilius:

## Description
Add minor image tags to allow automatic patch updates with [watchtower](https://github.com/containrrr/watchtower).

## How to test?
Add test command in Makefile:
`
test:
        @echo "test $(minor_tag)"
`

Run commands:
`
make test
make test tag=1
make test tag=1.22
make test tag=1.22.4
`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 .PHONY: build build-pg build-release build-arm build-x86 test clean
 
+minor_tag := $(shell echo $(tag) | cut -d. -f1,2)
 build-release:
 	@if [ -z "$(tag)" ]; then \
 		echo "Error: tag is required. Usage: make build-release tag=<tag>"; \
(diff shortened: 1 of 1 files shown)


Lokowitz:
What about also adding the major tag? If pangolin team agree to just make breaking changes to major releases, this could be beneficial for watchtower user too

oschwartz10612:
@Lokowitz am I understanding right that we do this currently? Usually major releases (1.X) are "breaking" and the minor versions (1.x.X) are patches - sometimes with some migrations.

@Shamilius Thanks for the PR! I think we would want to also update the CICD github action flow to generate and call the minor tag so its pushed to docker hub.

Shamilius:
@Lokowitz thanks for the feedback, I’ve added a major tag as well.  

@oschwartz10612 the Action runs `make`, so new tags are automatically pushed to the container registry. I’ve switched it to build all tags in one go - the Action should be a bit quicker as a result.

Lokowitz:
> @Lokowitz am I understanding right that we do this currently? Usually major releases (1.X) are "breaking" and the minor versions (1.x.X) are patches - sometimes with some migrations.

Yes you are currently doing right and just release new features with migrations (no need for user interaction) in minor releases.
But when the time comes that you need to release a breaking change (user interaction needed) it would be nice to have this in a major release.

Thanks :)

oschwartz10612:
Ahh yes okay I see. Yeah when we get to Pangolin 2.0 we will definitely 
bump that number haha! :}

oschwartz10612:
Thanks guys will test this in the next release!

PR #1580, opened 2025-09-28
Rules for rewriting requests to another path

Pallavikumarimdb:

## Description

**This PR introduces support for path rewriting rules.**

- Middleware creation for different rewrite strategies (exact, prefix, regex, stripPrefix).
- Extended config generation to include path rewrite middlewares in router chains.
- Ensures invalid rewrite configurations are logged and skipped.

This makes it possible to rewrite incoming request paths before forwarding to targets.
image

image

### DIFF
--- a/server/routers/traefik/getTraefikConfig.ts
+++ b/server/routers/traefik/getTraefikConfig.ts
@@ -108,12 +108,12 @@ function validatePathRewriteConfig(
         return { isValid: true };
     }
 
-    if ((rewritePath && !rewritePathType) || (!rewritePath && rewritePathType)) {
-        return {
-            isValid: false,
-            error: "Both rewritePath and rewritePathType must be specified together"
-        };
+    if (rewritePathType !== "stripPrefix") {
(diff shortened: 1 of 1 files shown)


Pallavikumarimdb:
Hi @oschwartz10612 , could you please review and let me know if any changes are needed?
@miloschwartz , can you also check if the UI requires any adjustments? Thanks!

oschwartz10612:
Couple small things: 

- [ ] I think if we are validating that we require the path to start with a / in the validatePathRewriteConfig we should also enforce that in the API, UI, and Blueprints. Alternatively we could prefix it with a / if it does not have one like on server/routers/traefik/getTraefikConfig.ts:617. I think same thing for stripPrefix
- [ ] When using the strip option and leaving the value blank to strip I get `[error]: Invalid path rewrite configuration for resource 1: Both rewritePath and rewritePathType must be specified together`

As far as I can tell everything else works! We love the UI!

Pallavikumarimdb:
@oschwartz10612 , I’ve made the requested changes. Could you please review and let me know if any further updates are needed?

oschwartz10612:
LGTM!! :}

PR #1586, opened 2025-09-29
Fix resource auth API call to use resourceGuid instead of resourceid

Pallavikumarimdb:

## Description
Resource Auth API call was incorrectly referencing `resource.resourceid` instead of the correct field `resource.resourceGuid`.

The mismatch (resourceid vs resourceGuid) caused the resource auth lookup to fail, triggering an unnecessary redirect back to
`/resources`


miloschwartz:
Thanks!

PR #1606, opened 2025-10-03
Feat/String interpolation in the headers.

Pallavikumarimdb:

## Description
Headers with like {{role}


oschwartz10612:
@Pallavikumarimdb I think what you implemented here was sending headers to the browser - not the downstream resource. I dont think we would want to send the headers to the browser but to the downstream resource so the resource could pull out the info.

But thinking through the implementation of that more now I think it would be a bit difficult because we are setting the headers in the traefik config currently instead of in the badger session. 

Thanks for the hard work on this though and sorry we are not taking it!

PR #1612, opened 2025-10-04
UI Adjustments

Pallavikumarimdb:

## Description
1. Add refresh button to all the pages.
2. Improve target configuration table design.
image

image

image


miloschwartz:
image

miloschwartz:
@Pallavikumarimdb Thanks! This is looking better. I got kind of carried away and implemented an advanced toggle to hide some of the columns by default, so it's easier to parse for new users. A couple of other things I noticed which would be good to worked out before merging:
- Instead of Dialog components we have a Credenza component which is essentially just a wrapper that is also mobile responsive. Can you check anywhere where there is a Dialog component and replace it with Credenza? The migration should be simple and you can take a look at other instances of the component to see how it's done.
- When creating a resource via the create wizard, there is a target table at the bottom so you can quick add targets. Can you look at standardizing this component so it matches the targets input table in the resource edit settings?
- Check for any untranslated strings in the these components and add them to the `en-us.json` file.

oschwartz10612:
LGTM! Thanks so much! :}

PR #1624, opened 2025-10-06
Feat/Cert Resolver Input Box

Pallavikumarimdb:

## Description
Input box with the domain to add cert resolver.
image
image
image

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -20,7 +20,7 @@ import { initTelemetryClient } from "./lib/telemetry.js";
 import { TraefikConfigManager } from "./lib/traefik/TraefikConfigManager.js";
 import { initCleanup } from "#dynamic/cleanup";
 import license from "#dynamic/license/license";
-
+import { fetchServerIp } from "./lib/serverIpService.js";
 async function startServers() {
     await setHostMeta();
 
@@ -30,6 +30,8 @@ async function startServers() {
     await license.check();
 
     await runSetupFunctions();
+
+    await fetchServerIp();
 
     initTelemetryClient();
 
--- a/server/lib/serverIpService.ts
+++ b/server/lib/serverIpService.ts
@@ -1,0 +1,28 @@
+import axios from "axios";
+
+let serverIp: string | null = null;
+
+const services = [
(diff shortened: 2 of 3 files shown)


oschwartz10612:
@Pallavikumarimdb the domain page is looking good! Thanks for working on this! Not sure if you are done but a couple of comments:

image
image
image

- [x] Can we make the type the same across the table and the page. Lets use the same language thats in the table like "Domain Delegation (NS)"
- [x] Instead of "unverified" can we match how it is in the table as "pending" and with a yellow badge. Also the same in the table of dnsRecords.
- [x] I like how you broke out the dns records into their own table. Maybe we could add a boolean column for verified status the same that the domains table has. Then you can pull the "pending"/"verified" status for the badges out of this table for each record and I can update the DNS servers to update the records themselves then they are verified.  
- [x] I think there is non-standard space between the title and description "Configure settings for your domain" and the info box with the type and the status. I also think there is nonstandard space between the required badge and "DNS records".

Pallavikumarimdb:
Hi @oschwartz10612 , thanks for the review! Yeah, it’s almost done - I’ll make the requested changes.

Pallavikumarimdb:
Hi @oschwartz10612 , I’ve made the change. Could you please review and let me know if any further adjustments are needed?

oschwartz10612:
I think this is looking really good! Thanks! We were messing around with it and had one more idea (sorry).

Could we:
- [x] Show the wildcard record info that's also shown when creating a wildcard type domain in the table on the domain page so all of the different domain types show records
- [x] Use ifconfig.io to query in the server on startup and store the IP of the server in memory. Then show the ip of the server in the DNS records for the wildcard domain type on the page for the records instead of `Server IP Address`. We think this will help people understand what to put in 
- [x] Remove the popup. We realized this does not make a ton of sense anymore because the info is shown on the edit page. So after creating the domain it can just take you to the page like when you create a resource / site / etc...

Pallavikumarimdb:
No worries at all! Glad you liked it. Sure, I’ll work on the changes you suggested.

PR #1625, opened 2025-10-06
Add docker dev image creation workflow for PRs

Lokowitz:
## Community Contribution License Agreement
By creating this pull request, I grant the project maintainers an unlimited,
perpetual license to use, modify, and redistribute these contributions under any terms they
choose, including both the AGPLv3 and the Fossorial Commercial license terms. I
represent that I have the right to grant this license for all contributed content.

## Description
This workflow creates an amd64 image for sqlite and postgresql version of each PR against `main` and `dev` branch in a docker hub repo specified with github variable `DOCKER_HUB_REPO`. 
This repo can be your original `fosrl/pangolin` or a new dev repo, maybe `fosrl/pangolin-dev`.
I removed the arm64 version because it needs ~40min instead of ~6min to create and maybe the amd64 version is enough for testing.

## Usecase
For faster and easier testing, especially on low power VPS. 
A low power VPS (1CPU,1gb RAM, 10gb DISK) is nearly unable to create the docker image.

image

Feel free to reject if you don't see a benefit 😊

PR #1661, opened 2025-10-13
Update dev-image

Lokowitz:

## Description
Sorry for the delay on fixing this.

> I have been trying to figure out why this is failing and I think it is because actions from PRs can't use secrets because they would be accessible and leaked.
> I removed it for now but am open to adding it back if we figure the right way to push.

Yes on `pull_request` is just working when someone from the team ist creating the PR, otherwiese no secrets will be available.
On `pull_request_target` secrects are available. But this workflow will then run with code from the PR target (trusted code).
So we have to merge in the new code with `checkout`. To be save that no one creates a PR which is stealing the secrets i have added an environment `build-dev`. This will bring up an approval for running the workflow.

It is not super nice to have this approval step but otherwise it will be to risky.

> [!IMPORTANT]
>  Be sure to first create the environment and select at least one reviewer before merging.
image

Fix https://github.com/fosrl/pangolin/pull/1625


Lokowitz:
But I am not sure if this is giving us the intended benefit. That would assume that you have to approve the workflow for each push... 
So i will leave this workflow in my fork, to get an image for my very low vps. 
Let me know if you want this with approval (as is in this pr), without approval (and maybe check if pr/push is from a team member, so that no failed workflow will be shown) or don't use this workflow. 

Cheers Marvin

Lokowitz:
@oschwartz10612 I updated the action to just trigger an approval when the PR is from a fork. So PRs from the same repo (including Dependabot) will trigger the image creation directly.

PR #1684, opened 2025-10-17
Make priority optional in schema

Pallavikumarimdb:

## Description
Update priority in target schema to sync with recent change in "hide path match and rewrite in raw resource"


oschwartz10612:
Yeah this is probably best. We noticed last minute it should be optional but didn't change the schema and just did a `|| 100` :joy:

Issue #1492, opened 2025-09-18
better-sqlite3 support


Lokowitz:
Hi guys,

i was trying to upgrade Pangolin to the latest better-sqlite3 version (currently 12.2.0).
The current version in this project is 11.7.0.
The latest possible upgrade is version 11.9.1 because since 11.10.0 they [disallowed to return a promise and asyc transactions].(https://github.com/WiseLibs/better-sqlite3/pull/1364)
Node v24 support is added in 12.0.0.

I was not able to find a good way to support both sync and async.

Any ideas how you want to make it future-proof? Maybe switching only to Postgresql oder back to normal sqlite3 instead of better-sqlite3?

oschwartz10612:
Thanks as always for helping us maintain these dependencies.

Sounds like to be future proof we need to fix the transactions? I think 
we want to keep supporting sqlite. I guess we would have to go around 
all of the code and make refactor so transactions dont return an async 
function.

Could be a bit of a refactor. Will need to figure out when to schedule 
in. What do you think?

Lokowitz:
Problem is that PG still needs async.
I tested a bit to support both, but you have to change a lot of files and it is not allowed to have both versions while compiling. 
The only way i got it to work was to split it in sqlite and pg files and delete the not needed files while building the image. 

I checked the [drizzle docs](https://orm.drizzle.team/docs/get-started-sqlite) and saw that they are also supporting libSQL. Maybe this is an option. 

I already tried to make a change to libSQL but there is also a lot of refactor work especially for all the migrations.

marcschaeferger:
IMPORTANT: We don't need to rush this. There is no need for a immediate implementation/migration!!!
@oschwartz10612 @Lokowitz 

One idea worth exploring and that ties directly into the async/sync discussion here is switching the SQLite driver in Drizzle from `better-sqlite3` to **libSQL**.

LibSQL is a fully open-source fork of SQLite ([GitHub repo](https://github.com/tursodatabase/libsql)) that Drizzle officially supports.  
For Node, there’s [libsql-js](https://github.com/tursodatabase/libsql-js), which offers a **better-sqlite3-compatible API** you can keep running in-process against a local `.db` file, while gaining:

- **Native async driver support** (avoids the async transaction limitation that blocks us past better-sqlite3 11.9.x)  
- **Easy switch to remote/cloud backends** like [Turso](https://github.com/tursodatabase/libsql)  
[...]

PR #1723, opened 2025-10-21
Switch DB from better-sqlite3 to libsql

Lokowitz:

## Description
Hey guys,
this is related to the [issue](https://github.com/fosrl/pangolin/issues/1492) that better-sqlite3 is not supporting async anymore.
This PR is an example how we could do the step to libsql with a minimum of effort.
This will require all users to migrate first to the latest available version (current 1.12.2) because I deleted all old migration files for sql because otherwise they need to be completely rewritten.
I added a check in `server/setup/migrationsSqlite.ts` to don't skip this task.

If you don't want to go this way and want to rewrite all migration files, than feel free to take over because this is out of my skills :-D 

Cheers Marvin


Lokowitz:
@oschwartz10612 fixed the merge conflicts. So feel free to review :)

oschwartz10612:
Thanks @Lokowitz!

Not opposed to switching to libsql necessarily. Forgetting if we discussed this earlier... is there a way we can keep better-sqlite3 and adjust the code to not complain about the async functions or that migration would be too difficult?

marcschaeferger:
@oschwartz10612 Yes, there are ways to keep `better-sqlite3` for example by refactoring everything to fully synchronous transactions or by pinning to 11.9.1 (which effectively means no proper Node 24 support). We could even introduce our own transaction wrapper for the async/sync differences. All of those options would technically work, but they either lock us into an older `better-sqlite3` version or require a non-trivial refactor and avoiding re-introducing async transaction callbacks. From a maintenance and future-proofing perspective, I still think adopting libSQL is the cleaner and less risky path.

What I would *not* do, though, is delete the old migration files. In theory we can keep all versions compatible and allow a direct upgrade from e.g. 1.10 → 1.13 without forcing users through 1.12.2 first. Drizzle will happily run all missing migrations in order as long as all the `scriptsSqlite/*.ts` files are still present **and** work with the new libSQL driver.

That does mean a bit more work (adapting and testing the legacy migrations), but in return we don’t artificially constrain users to a specific intermediate version just because we dropped older migrations from the codebase.

You can see some discussion in the following Issue https://github.com/fosrl/pangolin/issues/1492

oschwartz10612:
I think switching is a good plan then but I do agree with @marcschaeferger that we need to keep the migrations. I can work on porting them to libsql

Lokowitz:
@oschwartz10612 I added back all the migration files.
I am now on vacation for one week, so feel free to change the pr to your needs.

PR #1739, opened 2025-10-23
Add UI support for Proxy Protocol

Pallavikumarimdb:

## Description
Enable Proxy Protocol configuration through the UI for TCP and UDP resources.

### DIFF
--- a/install/config/traefik/dynamic_config.yml
+++ b/install/config/traefik/dynamic_config.yml
@@ -51,3 +51,12 @@ http:
       loadBalancer:
         servers:
           - url: "http://pangolin:3000"  # API/WebSocket server
+
+tcp:
+  serversTransports:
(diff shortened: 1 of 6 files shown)


oschwartz10612:
Thanks!

@Pallavikumarimdb apparently you cant define dynamic servers transport ref: https://github.com/traefik/traefik/issues/7893. This is news to me and one of those annoying Traefik quirks. 

I included two servers transports in the dynamic config file which we can add by default in new installs and though a migration for existing installs. One is for v1 proxy protocol and one for v2. Then in the getTraefikConfig functions instead of including the serversTransport sections you can just add the serversTransport reference in the service like you already have done.

Pallavikumarimdb:
@oschwartz10612 Got it, thanks for the clarification and fixing it! That makes sense now.

PR #1743, opened 2025-10-24
Feat/Option to regenerate Newt/olm/remote-nodes keys

Pallavikumarimdb:

## Description
Option to regenerate Newt/olm/remote-nodes keys (in case they're lost or need to be rotated)

### DIFF
--- a/messages/en-US.json
+++ b/messages/en-US.json
@@ -2111,7 +2111,7 @@
     "confirm": "Confirm",
     "regenerateCredentialsConfirmation": "Are you sure you want to regenerate the credentials?",
     "endpoint": "Endpoint",
-    "id": "Id",
+    "Id": "Id",
     "SecretKey": "Secret Key",
     "featureDisabledTooltip": "This feature is only available in the enterprise plan and require a license to use it."
 }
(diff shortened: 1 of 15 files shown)


oschwartz10612:
Looking good! A couple of comments. Meant to make these earlier but got lost in time sorry! 

- [ ] For local sites instead of disabling the button could we hide the whole credentials tab because there are no credentials in a local site
- [ ] Instead of "Regenerate Credentials" could we call it just "Re-key"
- [ ] Instead of doing it all in the card could we do it in a modal and make it a component so we could reuse if possible? I think the flow would be like this: click the button -> open up a modal asking the user to confirm that they want to rekey the site and explain that it this will mean that the old credentials will become invalid. Then when they click confirm thats when you save the new credentials into the database and show them to the user in a new modal. The the only button on the modal after they can see the credentials is to close the modal.
- [ ] Can you move the site and client rekeying into an enterprise feature as well? All you would need to do is move the API routes into the private API, then make sure to wrap the credentials page in the front end with a `build == "enterprise"` flag so it does not show on community, and also make sure we require a license to use it - so we should check that with the middleware in the api and make sure they cant click the button on the front end. You can use `verifyValidLicense` and `verifyValidSubscription` middleware on the routes and use the `useSubscriptionStatusContext` and `useLicenseStatusContext` on the front end

Pallavikumarimdb:
Hi @oschwartz10612 , I've made the requested changes. Could you please review again?

PR #1767, opened 2025-10-28
Make site and resource niceId editable

Pallavikumarimdb:

## Description

1. Make site and resource niceId editable
2. Hide site and resource "niceId" in the tables, and only show more info when clicking edit

### DIFF
--- a/server/routers/resource/updateResource.ts
+++ b/server/routers/resource/updateResource.ts
@@ -242,7 +242,12 @@ async function updateHttpResource(
         const [existingResource] = await db
             .select()
             .from(resources)
-            .where(eq(resources.niceId, updateData.niceId));
+            .where(
+            and(
+                eq(resources.niceId, updateData.niceId),
+                eq(resources.orgId, resource.orgId)
+            )
+        );
 
         if (
             existingResource &&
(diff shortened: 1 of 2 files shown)


oschwartz10612:
Hi @Pallavikumarimdb sorry about this. Having looked at the implementation when in the top card we think it will be better in the general form under the name as an input. Could you make that change?

We liked it at the top but decided it would be more application consistent in the general form on both the site and the resources.

image

Pallavikumarimdb:
@oschwartz10612 , I’ve made the change. Could you please review again?

server/routers/resource/updateResource.ts
```diff
@@ -362,6 +383,25 @@ async function updateRawResource(
 
     const updateData = parsedBody.data;
 
+    if (updateData.niceId) {
+        const [existingResource] = await db
+            .select()
+            .from(resources)
+            .where(eq(resources.niceId, updateData.niceId));
```
oschwartz10612: Here we should scope it to the orgId. Nice IDs only need to be unique within the same org.

Pallavikumarimdb:
Hi @oschwartz10612 , I've made the requested changes. Could you please review again?

PR #1789, opened 2025-11-01
Down for maintenance screen

Pallavikumarimdb:

## Description
Down for maintenance screen serve via the resource.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -454,50 +454,50 @@
             }
         },
         "node_modules/@aws-sdk/client-sesv2": {
-            "version": "3.910.0",
-            "resolved": "https://registry.npmjs.org/@aws-sdk/client-sesv2/-/client-sesv2-3.910.0.tgz",
-            "integrity": "sha512-YKCXrzbEhplwsvjAnMKUB9lAfawFHMz7tPLL3dCnKQYeZOQqsYiUUPjkiB2Vq8uV+ALZqJ2FCph07pW7XZHqjg==",
+            "version": "3.931.0",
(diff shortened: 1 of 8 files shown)


oschwartz10612:
Looking really good. I like all of the options for the maintenance page and the automatic vs forced mode is a good idea!

Couple of things: 

- [ ] I think we should move the settings into a new SettingsSection card under the general settings to separate it as an option. Use the normal pattern for the settings section with the title and everything I think. 
- [ ] I think we should remove the orange line on the left because we dont use that design pattern anywhere else.
- [ ] Looks like a bunch of the strings in the UI need to be translated and put into the en-US file
- [ ] I think I mentioned we want this to be an enterprise only feature. We would want to remove the maintenance mode stuff from the getTraefikConfig and keep it in the private getTraefikConfig. We should also wrap the new settings card in a `build != "oss"` check so it does not show on community. Then we should make sure it needs to be licensed - so setting the values in update resource should be rejected if it is not licensed and the toggle should be disabled if it is not licensed. You can show a alert if it is not licensed. Check out `src/app/[orgId]/settings/general/page.tsx` and the `updateOrg.ts` for how we are currently doing this.
- [ ] I think the maintenance server was a good idea to keep it separate from the nextjs server but I think it would be better to consolidate everything so the only UI we are using is nextjs. So I think we should remove the maintenance server and instead point the resource to the nextjs server in the traefik config. You could rewrite the path in this situation to go to something like `/maintenance-screen` or something in the next router. Then at that screen you could make a server component that gets the host header and then makes a request to the API to get the maintenance mode info from the DB and then render the page with react.

Sorry I know these are some big changes but I think this will be a great feature! :} 

image

oschwartz10612:
Thank you so much Pallavi!

PR #1792, opened 2025-11-01
Refactor CI/CD workflow for improved release process

marcschaeferger:
Updated CI/CD workflow to include new permissions, job definitions, and steps for version validation, tagging, and artifact management.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -4,9 +4,12 @@ name: CI/CD Pipeline
 # Actions are pinned to specific SHAs to reduce supply-chain risk. This workflow triggers on tag push events.
 
 permissions:
-  contents: read
-  packages: write      # for GHCR push
-  id-token: write      # for Cosign Keyless (OIDC) Signing
+  contents: write          # gh-release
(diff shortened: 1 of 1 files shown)


oschwartz10612:
@marcschaeferger I got a little too excited and merged this prematurely. A couple of things I think we should change:

- [ ] Its a good idea to build the server standalone and package it up as a tar. However I think we would need to do this 4x times: community-sqlite, community-postgres, enterprise-sqlite, enterprise-postgres.
- [ ] Similarly the docker buildx command needs to build 4x the containers its currently building: community-sqlite, community-postgres, enterprise-sqlite, enterprise-postgres. I think it currently is just doing it for community sqlite? This command should use the build args to control what to build. So `--build-arg BUILD=oss/enterprise` and `--buildarg DATABASE=sqlite/pg`

oschwartz10612:
- [ ] I think it needs to build the installer like the old CICD did and upload that as part of the release.
- [ ] Can we have it create a DRAFT release so we can still edit the release notes and check things before publishing?

If it already does some of this and I just don't under stand I am sorry!

PR #1809, opened 2025-11-04
Update German translations for client and blueprint terms

clemone210:

## Description
"Kunden" is generally used for "Customers", so in this case I suggest to stick with Client, as this is a widely used term in german tech sector. The same for "Bauplan" or "Blaupause". "Bauplan" is a "Construction plan" for building houses. "Blaupause" is pretty much the right translation for blueprints, but I would stick with Blueprint here as well.

## How to test?
Just translation proposal.


oschwartz10612:
Thanks!

jonasmerkel:
Hi @oschwartz10612 a few days ago I also translated from "Kunde" to "Client" however this translation got overwritten in another commit after it seems like some automatic translations were made. How can this be prevented?

oschwartz10612:
Hum this could be because of crowdin. It is possible crowdin is not 
properly syncing the translations from the repo. Ill look into this 
maybe I am doing something wrong.

It may be better to submit translation changes to the Crowdin project 
itself. https://crowdin.com/project/fossorial-pangolin

LMK if there are issues with this.

PR #1814, opened 2025-11-05
Feat: version updates & product updates popup

Fredkiss3:

## Description

**Main changes:**

- Added a new popup for updates and new version

**Other changes:**

- Added env variables validation and made it so that the fossorial api (`api.fossorial.com`) is replaced by an environment variable if it is provided.   
- The env variables are validated on local dev
- Moved `next.config.mjs` to `next.config.ts` to be able to import typescript files into it.

## Screenshots

| - | - |
| ------ | ------ |
| Video |  <video src="https://github.com/user-attachments/assets/a89ea530-1be8-4fbd-8142-9d9f96a54ef9"></video>  |


src/lib/queries.ts
```diff
@@ -0,0 +1,58 @@
+import { keepPreviousData, queryOptions } from "@tanstack/react-query";
+import { durationToMs } from "./durationToMs";
+import { build } from "@server/build";
+import { remote } from "./api";
+import type ResponseT from "@server/types/Response";
+
+export type ProductUpdate = {
+    link: string | null;
+    edition: "enterprise" | "community" | "cloud" | null;
```
miloschwartz: `oss` | `enterprise` | `saas`

Fredkiss3:
@miloschwartz 
> * When a link is applied to a specific product update, can we show an icon somewhere in that card so it's obvious to the user that it can be clicked to open the link. The use case here would be linking to docs, release notes, blog article, etc.c

- I forgot to add that, will do 👍🏾 

> * Just confirming, though I think I saw it in the code, do the product updates show in all builds (oss, enterprise, saas), and the new release only shows in oss and enterprise?

- Yes, see https://github.com/Fredkiss3/pangolin/blob/f27ae210ed7a2bdcfd1f1bdbcfc713192c4213d7/src/lib/queries.ts#L55-L56 

> * We would like to have two variables in the app section of the config file to disable product update notifications and a second one to disable release notifications.

- Noted, will do 👍🏾 

> * What happens if there are like 20 product updates? Does the popover just get really tall, or does it have a max height with a scroll bar? I think ideally there would be a max height with a scroll bar, but use your discretion for what looks best.

- Yes, there is a max height with a scrollbar that appear when there are a lot of updates, see: https://github.com/Fredkiss3/pangolin/blob/f27ae210ed7a2bdcfd1f1bdbcfc713192c4213d7/src/components/ProductUpdates.tsx#L220

> Lastly, this is super minor and just me being picky, but would it be possible to have this text extend further to the end of the card so we can fit a few more words in there?

- Noted, will do 👍🏾

Fredkiss3:
@miloschwartz 
I added two flags to the config for the notifcations: 

```yaml
app:
   # ... rest
   product_updates: true
   new_releases: true
```

PR #1819, opened 2025-11-06
Fix/Revert column from Resource table to fix type error and match overall styling

Pallavikumarimdb:

## Description
Revert column from Resource table


oschwartz10612:
Thanks for the catch - milo ended up doing that column control stuff on one of our branches because we forgot about your implementation so had to revert one apologies!

PR #1856, opened 2025-11-13
fix: Remove return in installer which prevents showing token

### LINKED ISSUE #1509: Setup token not displayed after installation
I installed pangolin using the installer as on the quick install guide and after the installation was completed the setup token was not displayed. However, the initial setup page on my domain still asks for the setup token. I suppose this is a bug.

LaurenceJJones:
Remove redundant return statement after success message.

## Description

Should fix #1509 after selecting to install crowdsec unless an error occurs. Its a quick fix, but you may want to move the showing of token before the crowdsec installation.

## How to test?

When installing pangolin and when accepting to use crowdsec should show setup code at the end of the installation.

### DIFF
--- a/install/main.go
+++ b/install/main.go
@@ -238,7 +238,6 @@ func main() {
 				}
 
 				fmt.Println("CrowdSec installed successfully!")
-				return
 			}
 		}
 	}


oschwartz10612:
Thanks @LaurenceJJones! I actually think after might make sense because the stack can get restarted from Crowdsec and the token would change I think. But if this causes issues we could move it before.

PR #1932, opened 2025-11-27
Fix resource table card header responsive issues

filipe-costa:

## Description

When viewing the "Manage Resources" page on mobile and when they are available clients, i.e. `enableClients` flag is true - the `Proxy` and `Client` resources tab list buttons would end up overlapping itself.

I took the opportunity to use CSS Grid instead of flex box since I wanted to have more granular control over both axis - and at different breakpoints.

I did a local docker build in my own VPS environment to showcase this:

Before the fix - I have tested it with Chrome  on Iphone 13 Pro:

After the fix  - I have tested it with Chrome on Iphone 13 Pro:

I have also tested in the following mobile browsers:
- Firefox
- Safari

## How to test?
- In `src/lib/pullEnv.ts`, and set the key `enableClients` to be `true`.
- Use different browsers on Desktop using the Responsive Design Mode at different breakpoints.

### DIFF
--- a/src/components/ResourcesTable.tsx
+++ b/src/components/ResourcesTable.tsx
@@ -1,79 +1,73 @@
 "use client";
 
-import {
-    ColumnDef,
-    flexRender,
-    getCoreRowModel,
-    useReactTable,
-    getPaginationRowModel,
-    SortingState,
-    getSortedRowModel,
-    ColumnFiltersState,
-    getFilteredRowModel,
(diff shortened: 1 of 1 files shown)


miloschwartz:
Hey, thanks for working on this! Closing because we split the tables into two and have therefore removed the tab selector you have in your screenshots. Therefore this fix isn't needed anymore.

PR #1971, opened 2025-12-04
feat(healthcheck): add SNI support for target healthchecks

water-sucks:

## Description

While an SNI proxy already exists for configured targets themselves, this is not the case for sending periodic requests to targets' healthcheck endpoints.

This PR adds a field to the target healthcheck settings to specify an explicit SNI header for target healthcheck requests.

Screenshot 2025-12-03 at 23 27 49

Merging this PR requires https://github.com/fosrl/newt/pull/185 to be merged alongside it; as such, I'm not sure how the process goes for coordinating new behavior between Pangolin and Newt, but I believe that older versions of Newt would simply not support this field and silently ignore it since it won't get parsed out of the JSON response.

## How to test?

Fill in the SNI field for a given target healthcheck and make sure that the downstream provider receives/proxies the client hello correctly as specified by the configured target healthcheck settings.

### DIFF
--- a/messages/en-US.json
+++ b/messages/en-US.json
@@ -526,6 +526,8 @@
     "targetCreatedDescription": "Target has been created successfully",
     "targetErrorCreate": "Failed to create target",
     "targetErrorCreateDescription": "An error occurred while creating the target",
+    "tlsServerName": "TLS Server Name",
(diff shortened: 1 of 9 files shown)


oschwartz10612:
Thanks!

PR #2045, opened 2025-12-11
fix: full domain should be updated when the form is saved on resource proxy page

Fredkiss3:

## Description

I noticed a bug where when updating a resource domain, it would not be reflected in the infobox, and when going back and forth to the homepage, the value of the full domain would revert to the previous value. 

The domain is correctly updated on the backend though, since we can see that if we refresh the page, it would reflect the actual value we updated. 

The fix: calling `updateResource()` with the updated value, not the previous value of the resource. 

## screenshots 
<table>
<tr>
 <td align="center">
  video
 </td>
</tr>

<tr>
 <td>
<video src="https://github.com/user-attachments/assets/ae0fff1f-dfe1-4fdb-9325-15f620ca0a7d"></video>
 </td>
</tr>
</table>


oschwartz10612:
Thanks!

PR #2084, opened 2025-12-15
ci: parallelize test workflow

water-sucks:

## Description

To enable a faster feedback loop when pushing up to a pull request, move the Docker image building for both PostgreSQL and SQLite into separate CI jobs.

Additionally, the workflow itself seemed broken. `make build-sqlite` uses `buildx` and requires a `tag` param now, which is causing testing CI workflows to fail when they are not able to build the Docker image.

Older CI workflows used to run `docker build` without the `buildx` driver, and the rule that does this got renamed to `dev-build-${db}`, so I switched the rule name to get this working again.

## How to test?

The `test.yml` workflow should pass and be substantially faster.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,11 +12,12 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
-
     steps:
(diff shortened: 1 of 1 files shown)


oschwartz10612:
:heart:

PR #2133, opened 2025-12-21
Fix: Prevent cache memory leak with maxKeys limit and conditional caching

### LINKED ISSUE #2120: Pangolin leaking memory after upgrading to 1.13.1
### Describe the Bug

Pangolin seems to be leaking memory after upgrading from 1.10.2(?) to 1.13.1. 

Image

### Environment

- OS Type & Version: Ubuntu 24.04.3 LTS
- Pangolin Version: 1.13.1
- Gerbil Version: 1.3.0
- Traefik Version: 3.6

### To Reproduce

~8k allowed and ~2k blocked requests a day.

### Expected Behavior

Memory usage stays constant.

djcrafts:
### Summary
Fixes #2120 by preventing unbounded cache growth through three key improvements:
1. Adding a `maxKeys` limit to the NodeCache instance
2. Skipping cache writes for unsuccessful lookups
3. Adding periodic cache statistics logging for monitoring

### Problem
After upgrading to 1.13.1, Pangolin was experiencing memory leaks, especially when GeoIP (MaxMind) database was configured. With ~10k requests per day, the cache would grow unbounded, eventually causing OOM kills on memory-constrained systems.

**Root causes:**
1. NodeCache had no `maxKeys` limit, allowing unlimited growth
2. GeoIP/ASN lookups were cached even when they returned `undefined` (e.g., when MaxMind DB not configured)
3. No visibility into cache performance made the issue hard to diagnose

#### 1. Add maxKeys limit (cache.ts)
```typescript
export const cache = new NodeCache({
    stdTTL: 3600,
    checkperiod: 120,
    maxKeys: 10000  // Prevents unbounded growth
});
```
- Sets maximum of 10,000 cached keys
- Uses LRU eviction when limit is reached
- With ~10k requests/day and 5min TTL, provides ample headroom

#### 2. Skip caching undefined values
**Before:**
```typescript
cachedCountryCode = await getCountryCodeForIp(ip);
cache.set(geoIpCacheKey, cachedCountryCode, 300); // Caches undefined too
```

**After:**
```typescript
cachedCountryCode = await getCountryCodeForIp(ip);
if (cachedCountryCode) {  // Only cache successful lookups
    cache.set(geoIpCacheKey, cachedCountryCode, 300);
}
```

#### 3. Add cache monitoring
```typescript
setInterval(() => {
    const stats = cache.getStats();
    logger.debug(
        `Cache stats - Keys: ${stats.keys}, Hits: ${stats.hits}, Misses: ${stats.misses}, Hit rate: ...`
    );
}, 300000); // Every 5 minutes
```

### Testing
- ✅ TypeScript compilation successful (no type errors)
- ✅ No functional changes to lookup logic
- ✅ Backward compatible with existing cache usage
- ✅ Cache statistics will be visible in debug logs

### Impact
- **Memory usage**: Capped at ~10k keys instead of unbounded growth
- **Performance**: No degradation; LRU ensures frequently accessed IPs remain cached
- **Monitoring**: Cache stats in logs help identify issues early
- **Undefined caching**: Prevents wasting cache space on failed lookups

### Related
- Addresses issue raised by @Ragnaruk about lack of cache stats logging
- Addresses issue raised by @nath1416 about OOM kills on 1GB VPS
- ASN lookup caching already had conditional caching; now GeoIP matches that pattern

### DIFF
--- a/server/lib/cache.ts
+++ b/server/lib/cache.ts
@@ -1,5 +1,20 @@
 import NodeCache from "node-cache";
+import logger from "@server/logger";
 
-export const cache = new NodeCache({ stdTTL: 3600, checkperiod: 120 });
+// Create cache with maxKeys limit to prevent memory leaks
+// With ~10k requests/day and 5min TTL, 10k keys should be more than sufficient
+export const cache = new NodeCache({
+    stdTTL: 3600,
+    checkperiod: 120,
+    maxKeys: 10000
+});
+
+// Log cache statistics periodically for monitoring
+setInterval(() => {
+    const stats = cache.getStats();
+    logger.debug(
+        `Cache stats - Keys: ${stats.keys}, Hits: ${stats.hits}, Misses: ${stats.misses}, Hit rate: ${stats.hits > 0 ? ((stats.hits / (stats.hits + stats.misses)) * 100).toFixed(2) : 0}%`
+    );
+}, 300000); // Every 5 minutes
 
 export default cache;
--- a/server/private/lib/logAccessAudit.ts
+++ b/server/private/lib/logAccessAudit.ts
@@ -161,8 +161,11 @@ async function getCountryCodeFromIp(ip: string): Promise<string | undefined> {
 
     if (!cachedCountryCode) {
         cachedCountryCode = await getCountryCodeForIp(ip); // do it locally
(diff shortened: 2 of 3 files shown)


oschwartz10612:
Thank you! :smile:

PR #2148, opened 2025-12-22
fix: Don't treat newt release-candidate as a "update" in the site list

huzky-v:

## Description
As of 2025-12-22, the current stable version of newt is `1.7.0`, but the UI shows there is an available update. The *newest* version is `1.8.0-rc.0`. And I think it will be better to show the update icon when there is a newest stable update, but not a release-candidate update.

The behavior does not align with `https://static.pangolin.net/get-newt.sh` as the github API checked in newt install and the site API is different. 
Newt install script: `https://api.github.com/repos/fosrl/newt/releases/latest`
Site API: `https://api.github.com/repos/fosrl/newt/tags`

Or it will be better to store the latest version elsewhere to avoid confusion.

### DIFF
--- a/server/routers/client/listClients.ts
+++ b/server/routers/client/listClients.ts
@@ -56,12 +56,12 @@ async function getLatestOlmVersion(): Promise<string | null> {
             return null;
         }
 
-        const tags = await response.json();
+        let tags = await response.json();
         if (!Array.isArray(tags) || tags.length === 0) {
             logger.warn("No tags found for Olm repository");
             return null;
         }
-
+        tags = tags.filter((version) => !version.name.includes("rc"));
         const latestVersion = tags[0].name;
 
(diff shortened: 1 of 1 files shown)


oschwartz10612:
@huzky-v Agreed and thank you for the fix! Could you also apply this fix to the olm version check and the remote node check. They operate the same way and will not ignore the release candidates.

huzky-v:
Just done the olm part

And I am not sure about the remote node part, as I found just olm and newt after searching the pangolin codebase.
listSites: https://api.github.com/repos/fosrl/newt/tags
listClients: https://api.github.com/repos/fosrl/olm/tags

oschwartz10612:
Whoops you are right there is no update for that right now. Thanks for implementing olm so fast! :smile:

PR #2095, opened 2025-12-16
Add ASN-based resource rule matching

WildeTechSolutions:
- Add MaxMind ASN database integration
- Implement ASN lookup and matching in resource rule verification
- Add curated list of 100+ major ASNs (cloud, ISP, CDN, mobile carriers)
- Add ASN dropdown selector in resource rules UI with search functionality
- Support custom ASN input for unlisted ASNs
- Add 'ALL ASNs' special case handling (AS0)
- Cache ASN lookups with 5-minute TTL for performance
- Update validation schemas to support ASN match type

This allows administrators to create resource access rules based on Autonomous System Numbers, similar to existing country-based rules. Useful for restricting access by ISP, cloud provider, or mobile carrier.

## Community Contribution License Agreement
By creating this pull request, I grant the project maintainers an unlimited,
perpetual license to use, modify, and redistribute these contributions under any terms they
choose, including both the AGPLv3 and the Fossorial Commercial license terms. I
represent that I have the right to grant this license for all contributed content.

## Description
The implementation is nearly identical to using the Country rules with the maxmind GeoIP database.   This time you just need the `GeoLite2-ASN.mmdb` next to your `GeoLite2-Country.mmdb`.  Fetching the ASN database is no different from fetching the Country database, and if Pangolin detects it, it will enable the rule capability just like with Country.

image

image

Similar to country, a number of popular ASNs are provided in the dropdown by default.  This is a great way to reduce unwanted traffic to resources that can't support other Authentication means (i.e. Mobile apps that need api access that don't support custom headers like Amperfy).  

## How to test?
Add the `geoip-updater` container to `docker-compose.yml`.

```yml
  geoip-updater:
    image: maxmindinc/geoipupdate:latest
    container_name: geoip-updater
    volumes:
      - ./config/maxmind:/usr/share/GeoIP
    environment:
      GEOIPUPDATE_ACCOUNT_ID: ${GEOIP_ACCOUNT_ID}
      GEOIPUPDATE_LICENSE_KEY: ${GEOIP_LICENSE_KEY}
      GEOIPUPDATE_EDITION_IDS: ${GEOIP_EDITION_IDS}
    restart: on-failure
```

The `GEOIP_EDITION_IDS` variable should contain the ASN database:

```properties
GEOIP_EDITION_IDS=GeoLite2-Country GeoLite2-City GeoLite2-ASN
```

Add the `maxmind_asn_path` variable to the `server` block in `config.yml`:

```yml
server:
[...]

PR #2149, opened 2025-12-22
Region-based Resource Rules

infiniteWays:

## Description
Hi Owen and Pangolins!

I just saw the pull request #2095 and was intrigued to add also a region rule. 
I realized that something like that would be useful for regions of the earth with a high country density, it can be extremely tedious to create large amounts of rules.

I had a look around on what regional "classifications" there are and chose the one of the [United Nations Statistical Division ("M49 Standard")](https://unstats.un.org/unsd/methodology/m49) which apparently also somehow got into [Unicode](https://www.unicode.org/cldr/charts/48/supplemental/territory_containment_un_m_49.html).

This feature aims to add the rule cateogory "Region" based on that.
image
image
The numbers in parentheses are the M49 codes according to the UNSD.

Also included are three language localizations that I could do. English, German and Russian. I assume one could use a regular translator for the rest as the localizations are just single words. I however didn't want to mess with any "translation workflow", so I skipped on language I do not speak good enough.

Happy Holidays,
Dennis

## How to test?
Testing this should work straight without any modifications to the way you run the container. 
It is based on the same principles as the country rules.

I had some issues with integration testing resource accesses locally. If anyone has some time, a nice tutorial would be awesome!
Also, I had some issues with the drizzle.config.ts files in WSL/Windows. I had to exchange the `const schema = [path.join(...)]` with `const schema = ["server/db/sqlite/schema/"]`. Not sure if this is good or not, but it made me follow the instructions in the docs.

I however verified with curl that verifySession responds correctly according to the set rules shown above when messing around with the reported IP:
``` 
curl -X POST http://localhost:3001/api/v1/badger/verify-session -H "Content-Type: application/json" -d '{
"originalRequestURL": "http://test.example.com/api/test",
"scheme": "http",
"host": "test.example.com", 
"path": "/api/test",
"method": "GET", 
"tls": false, 
"requestIp": "{IP_of_your_choice}", 
"sessions": {}, 
"headers": {}, 
"query": {}  
}'
```
reports back 
```
{"data":{"valid":true},"success":true,"error":false,"message":"Access allowed","status":200}
```
while exchanging the IP with an IP outside Europe, the API reports:
```
{"data":{"valid":false},"success":true,"error":false,"message":"Access denied","status":200}
```

Please thoroughly double-check this on a working integration test server.

## What could be improved?

- Have the filter allow typing in countries and show their appropriate region. So if you type "Germany", show "Europe" and "Western Europe" in the dropdown
- Maybe add "Country Groupings" like the EU, the Euro Zone, etc. But I would assume this to be another filter rule in general.


Enzo1603:
"Hi Dennis,

I was just about to open a feature request for regional rules when I saw your PR.

I really like the idea of grouping countries into regions to keep the rule list clean. While looking at your screenshots, I had an additional thought on the UI: What do you (and the maintainers) think about a hierarchical multi-select for the 'Value' field?

I’ve attached a very basic mockup to illustrate the principle. In the mockup, I used a Continent > Country structure, but based on your PR, we could easily keep Region > Country  (or even extend it to Continent > Region > Country):
Pangolin-Country-Multiselect

The idea behind this:

- It would allow selecting multiple countries or even entire regions within a single rule, which would prevent 'rule sprawl' even further (right now, i.e. without your PR,  we can only select all countries and then make exceptions for individual countries).

- It makes handling exceptions very easy: You could select a whole region and then just uncheck one or two specific countries.

Just wanted to share this here as a suggestion. I think combining your regional logic with a multi-select approach would be great.

oschwartz10612:
Thanks this is great!

PR #2209, opened 2026-01-04
[Enhancement]: Be able to define priorities in blueprint rules configurations.

JanGrosse:

## Description
Currently when defining blueprints for e.g. public-resources one can define rules (https://docs.pangolin.net/manage/blueprints#rules-configuration). But as of now those rules do not include the `priority` setting. 
This PR is trying to change that by adding them as optional field to not break existing blueprints.

## How to test?
Spin up the testing environment, add a new blueprint through the GUI. 
For example
```yaml
public-resources:
  demo:
    name: demo
    protocol: http
    full-domain: demo.example.com
    host-header: example.com
    tls-server-name: example.com
    rules:
      - action: deny
        match: ip
        value: 1.2.3.1
        priority: 100
      - action: pass
        match: ip
        value: 1.2.3.2
        priority: 200
```
Check the ressources tab, it should now include the defined priorities 100 & 200 instead of the auto-assigned values 1 & 2.
grafik

## Additional Information
It also validates the rule that priorities should not be duplicated (same as when adding them manually). 
Additionally it takes into account that priorities get auto-assigned a value based on the index of the rule + 1.

So when defining a rule with a mix of assigned and unassigned priorities it "simulates" the auto-assignment during validation to check for duplicated priorities which can occur in mixing. 
For example the following rules would be invalid as the first rule gets auto-assigned the priority 1 and the follwoing rule tries to assign the same priority.
```yaml
[...]
    rules:
      - action: deny
        match: ip
        value: 1.2.3.1
      - action: pass
        match: ip
        value: 1.2.3.2
        priority: 1 <--- Invalid
```

### DIFF
--- a/server/lib/blueprints/proxyResources.ts
+++ b/server/lib/blueprints/proxyResources.ts
@@ -587,13 +587,15 @@ export async function updateProxyResources(
 
             // Sync rules
             for (const [index, rule] of resourceData.rules?.entries() || []) {
+                const intendedPriority = rule.priority ?? index + 1;
                 const existingRule = existingRules[index];
                 if (existingRule) {
                     if (
                         existingRule.action !== getRuleAction(rule.action) ||
                         existingRule.match !== rule.match.toUpperCase() ||
                         existingRule.value !==
-                            getRuleValue(rule.match.toUpperCase(), rule.value)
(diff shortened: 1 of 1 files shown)


oschwartz10612:
LGTM thank you!

PR #2245, opened 2026-01-14
feat(fingerprint): add endpoint to reset olm secret based on fingerprint 

water-sucks:

## Description

This attempts to reset an olm secret if a corresponding secret is found.

## How to test?

Re-login with a machine after logging out with a client, AND make sure the `olm` credentials are deleted client-side.

Any clients that use this endpoint should be able to reuse the same `olm` with a secret that has been reset (aka not the old secret, this should be removed), and a new device should not pop up in the list.

### DIFF
--- a/server/routers/external.ts
+++ b/server/routers/external.ts
@@ -861,6 +861,12 @@ authenticated.get(
     olm.getUserOlm
 );
 
+authenticated.post(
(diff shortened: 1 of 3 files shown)


oschwartz10612:
LGTM pretty straight forward. Thanks!

PR #2306, opened 2026-01-23
fix: tab between host & port in resource target address column 

Fredkiss3:

## Summary

- Fix tabulation between host and port inputs in target address in resources by memoizing the `getColumns` function and all its dependencies  
- Refactored the target address component into its own component
- Also fixed the focus state of the components in the target address to have the same visual state

## Screenshots

| Before               | After |
| :----------------: | :--------: |
| <video src="https://github.com/user-attachments/assets/a7acc694-51ad-4a28-88bf-ce8623d6f084"></video> | <video src="https://github.com/user-attachments/assets/51b3ad48-03d6-4731-8abc-b3fbb14aadcf"></video> |


miloschwartz:
This has been annoying me so much!

PR #2330, opened 2026-01-25
Remove unused packages and simplify files

Lokowitz:

## Description
1. removed unused packages and reduced image size from 1,35G to 1,26G
2. simplified Dockerfile by using npm run 
3. fix ARG issue by moving the ARGs for the label to the runner stage
4. removed dep type prod and dev from gomod Dependabot because they are not available for gomod

Maybe there is more potential to reduce image size, by removing redundant packages, for example:
- js-yaml -> yaml
- moment -> date-fns

Will check this later


Lokowitz:
Did some shit. Need to read more docs first 🥲

PR #2397, opened 2026-02-02
feat: Add support for rootless Podman execution

arousalspoon204:

## Description
This PR introduces support for running the Pangolin stack in rootless Podman mode and allows users to configure custom HTTP/HTTPS ports during installation.

### Key Changes:
 * **Rootless Support:** Added detection for rootless environments and adjusted container capabilities (conditionally disabling `SYS_MODULE` when rootless).
 * **Configurable Ports:** Users can now specify HTTP and HTTPS ports during the installation process, which are then correctly applied to the `docker-compose.yml
 * **Host Prerequisite Checks:** The installer now proactively checks for:
     * IP Forwarding (required for VPN routing).
     * `wireguard` kernel module (required on the host for rootless Gerbil).
     * `/dev/net/tun` access permissions.
     * Unprivileged port configuration (`ip_unprivileged_port_start`) for Podman.
 * **UX Improvements:** Added sudo-integration for host configuration tasks if the installer is run as a non-root user.

These changes make Pangolin more flexible for deployment on systems where root access is restricted or where standard ports (80/443) are already in use.

## How to test?
 1. **Run as non-root:** Execute the installer without sudo: `go run install/main.go`
 2. **Select Podman:** Choose Podman when prompted.
 3. **Test Ports (Two Scenarios):**
    *   **Scenario A (Standard Ports):** Enter `80` / `443`. Verify that the installer detects this and asks to configure `net.ipv4.ip_unprivileged_port_start` so
     Podman can bind these low ports.
    *   **Scenario B (Custom Ports):** Enter `8080` / `8443`. Verify that the generated `docker-compose.yml` uses these ports instead of the defaults.
 4. **Verify Compose:** Check the generated `docker-compose.yml` to ensure `SYS_MODULE` is omitted in rootless mode.

### DIFF
--- a/install/main.go
+++ b/install/main.go
@@ -21,9 +21,9 @@ import (
 
 // DO NOT EDIT THIS FUNCTION; IT MATCHED BY REGEX IN CICD
 func loadVersions(config *Config) {
-	config.PangolinVersion = "replaceme"
-	config.GerbilVersion = "replaceme"
-	config.BadgerVersion = "replaceme"
+	config.PangolinVersion = "latest"
+	config.GerbilVersion = "latest"
+	config.BadgerVersion = "v1.3.1"
 }
 
 //go:embed config/*
@@ -50,6 +50,9 @@ type Config struct {
 	EnableGeoblocking         bool
 	Secret                    string
 	IsEnterprise              bool
+	HttpPort                  int
+	HttpsPort                 int
+	Rootless                  bool
 }
 
 type SupportedContainer string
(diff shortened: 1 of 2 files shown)


install/main.go
```diff
@@ -22,9 +21,9 @@ import (
 
 // DO NOT EDIT THIS FUNCTION; IT MATCHED BY REGEX IN CICD
 func loadVersions(config *Config) {
```
LaurenceJJones: revert lines as per comment above function call

oschwartz10612:
I think I would prefer this to be a guide rather than built into the installer.

arousalspoon204:
I'm not a developer, but wouldn't it make sense to set up Pangolin with Podman as rootless by default, instead of with full root privileges as before?

In my opinion, the effort required for a hardened setup should not be left to the user. I tried to convert an existing installation to a rootless structure myself, but it was extremely tedious and, unfortunately, did not work stably in the end or ended in chaos.

Would an installer flag be a possible compromise? This would allow users to opt in for additional security without having to manually rebuild the system.

PR #2398, opened 2026-02-02
fix dockerfile creating non oss builds

Lokowitz:

## Description
There is a bug in building non oss build because i missed the -b tag in esbuild.
Tried to use the server/build.ts const but import is not working so i asked ai.
I am sure you know a better way to extract the build value from there.

If you want i revert this and add 3 npm run commands instead.

I am currently testing Docker Hardened Image. Needs some more refactoring, are you interested?
https://github.com/Lokowitz/pangolin/pull/551


oschwartz10612:
Thanks!

PR #2434, opened 2026-02-08
feat: Add path setting to share links (resourceAccessToken)

NHClaessens:

## Description
Adds an optional `path` parameter to share links that allows specifying a custom redirect path when accessing a resource.

This feature addresses a specific use case, if this change is not desired in the main branch, or if any changes are required, I would be more than happy to cooperate.

The use case in question:
- I run an Immich instance with shared albums
- In order to allow users to upload pictures, the `/api/` endpoint needs to be exposed, which is undesirable
- Share links would mitigate this, as they use a cookie to only allow users with a link to access the resource
- However, the link redirects to the base URL of the resource, with no option to change this
- Thus, I have added the option to add an (optional) path onto a share link's redirect (e.g. `share/xxxx`)

## Changes Made
- Added `path` column in the `resourceAccessToken` schema in Postgres and SQLite schemas (optional field, so not a breaking change)
- Added field in CreateSharedLinkForm.tsx
- Change generateAccessToken.ts to accept and return `path` value
- Change authWithAccessToken.ts to add the path onto the redirect url
- Add translations for the new field

## How to test?
1. Go to `Access` > `Links`
2. Create a new Share Link
3. Optionally add a path
4. Confirm that the final redirect leads to the expected location, and does not result in malformed URLs

(This is my first open-source contribution, thank you for your patience)


noahbroyles:
This feature would be amazing!

oschwartz10612 (approved):

oschwartz10612:
Thanks so much!

PR #2491, opened 2026-02-17
fix(install): add error handling, code cleanups, and YAML type refactor

rodneyosodo:

## Description
Improves the install package with better error handling and code quality cleanups across multiple files. Replaces `interface{}` with `any` for YAML map types to use modern Go idioms.

## How to test?

1. Run the installer and verify it completes without errors.
2. Test error paths (e.g., invalid config, missing input) to confirm errors are surfaced correctly instead of silently ignored.
3. Confirm CrowdSec and container setup still behave as expected.

### DIFF
--- a/install/config.go
+++ b/install/config.go
@@ -118,19 +118,19 @@ func copyDockerService(sourceFile, destFile, serviceName string) error {
 	}
 
 	// Parse source Docker Compose YAML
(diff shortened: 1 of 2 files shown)


oschwartz10612:
Thanks so much!

PR #2493, opened 2026-02-17
feat(resources): add redirect domain management in settings

LaurenceJJones:

## Description

I’m not sure of the best migration approach here, since this feature requires adding a new column to the resources table so its bundle inside of the PR, but maintainers may want to change it to a minor version since its a new feature?

however, there may be another option which we create a dedicate table for this? 🤷🏻 

  - add redirect domains section in resource settings
  - add "Add Redirect" flow using DomainPicker
  - validate redirect domains to prevent primary-domain and duplicate entries
  - normalize redirect domains before save (ASCII/lowercase)
  - support removing redirect domains from the configured list
  - add translation-key fallbacks for new redirect copy

## How to test?

When editing a resource, under the general tab you now have "Redirect" options which allow you to set subdomains or another domain (that added in the system) to redirect to the currently edited resource. This is useful as users may host a website on the base domain and requests to `www.` should be redirected to the base domain on a permanent redirect (or the reverse depending on the user preference).

image

Before scaffolding this feature the only option would be to create another resource that take the subdomain and points to the same target, however, that is not scalable and creates duplicate records.

Open to comments/feedback on the PR as direction may have to change or may prompt new ideas on how to achieve this.

### DIFF
--- a/server/db/pg/schema/schema.ts
+++ b/server/db/pg/schema/schema.ts
@@ -141,7 +141,8 @@ export const resources = pgTable("resources", {
     maintenanceTitle: text("maintenanceTitle"),
     maintenanceMessage: text("maintenanceMessage"),
     maintenanceEstimatedTime: text("maintenanceEstimatedTime"),
-    postAuthPath: text("postAuthPath")
+    postAuthPath: text("postAuthPath"),
+    redirectDomains: text("redirectDomains")
 });
 
 export const targets = pgTable("targets", {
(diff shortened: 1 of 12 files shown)


LaurenceJJones:
Thinking more about the feature, it may be better design decision to think of these as "HTTPS" resources which could either point towards another resource so for example in the "target" section instead of a site + url it could just be an extension then there is more power in the resources section as nothing is hidden under the edit tab of a resource?

LaurenceJJones:
Closing due to conflicts and also maintainers want a different direction EG: a table that has redirect to primaries so they can he handled differently within traefik configuration as this current design if user went http to redirect they would go http -> https redirect domain -> https primary, we should skip the http to https for redirect and direct directly to primary https

Can reopen if mostly right but needs joining table 🤷🏻

PR #2499, opened 2026-02-18
enhance(installer): use ldflags to inject versions

LaurenceJJones:

## Description

Instead of the CI/CD using sed to replace the 'replaceme' text we can instead use ldflags which can inject variables at build time to the versions. The makefile had a bunch of workarounds for dev so these have been removed to cleanup etc etc and fetchs versions from the gh api directly if the variables are not injected like the CI/CD does

## How to test?

Using `make` or explicit `make go-build-release PANGOLIN_VERSION=1.15.4 GERBIL_VERSION=1.0.0 BADGER_VERSION=1.0.0` will inject the variables into the binary without using `sed`

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -289,22 +289,14 @@ jobs:
                   echo "LATEST_BADGER_TAG=$LATEST_TAG" >> $GITHUB_ENV
               shell: bash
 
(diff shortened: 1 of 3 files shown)


oschwartz10612:
Thank you this is great!

PR #2500, opened 2026-02-18
enhance(installer): use huh package to handle input

LaurenceJJones:

## Description

Instead of relying on stdin and stdout by default, using the huh package from charmbracelet allows us to handle user input more gracefully such as y/n instead of typing 'yes' or 'no'. If a user makes a mistake whilst typing in any text fields they cannot use left or right to edit a single character when using huh it can. This adds a dependancy and may increase the size of installer but overall improves user experience.

## How to test?

Building the installer now uses components from huh package to handle just user input (instead of building a full weight TUI which is too much for just an installer). You can simply run the installer once built and to test the fallback logic you can run `TERM=dumb ./installer`

render1771428439682


oschwartz10612:
Thanks so pretty :eyes:

PR #2509, opened 2026-02-20
Podman Quadlet: Network=host for newt

curiousleo:

## Description
`newt` needs to run on the `host` network in order to communicate with local services exposed at 127.0.0.1 or localhost.

## How to test?
Try running the Podman Quadlet `newt` without this setting -- it won't be able to connect to local services.

### DIFF
--- a/src/components/newt-install-commands.tsx
+++ b/src/components/newt-install-commands.tsx
@@ -114,6 +114,7 @@ Description=Newt container
 [Container]
 ContainerName=newt
 Image=docker.io/fosrl/newt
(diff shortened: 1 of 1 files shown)


curiousleo:
I'm certainly no expert here, but this change definitely made things work for me.

oschwartz10612:
Thanks I think this is a fine default.

oschwartz10612:
Actually on second thought we do not do this for compose so I dont want to confuse people by doing this on podman. I

PR #2522, opened 2026-02-23
fix: logo URL validation shows cryptic error instead of meaningful message

### LINKED ISSUE #2325: Too Big Error Updating Logo Url
### Describe the Bug

Thanks for fixing the previous bug setting a url for the logo. 

The previous url I was using worked. This url structure does not.

Change it and its throwing an error using a url structure like this -
URL - https://subdomain.domainname.com/images/Microsoft1.png - example. This is an active url - changed the values since I didnt want to post my actual url but provides a view of the structure which isn't anything complicated.

Error -

`Too big: expected string to have <=0 characters`

### Environment

- Pangolin Version: 1.15.1
- Traefik Version: 3.6.7

### To Reproduce

Enter logo url in the form of - URL - https://subdomain.domainname.com/images/Microsoft1.png - example. This url wont work since I didnt want to post my actual url.

### Expected Behavior

For the logo url to be set.

shreyaspapi:

## Summary

Fixes #2325

- Replaced `z.union([z.literal(""), z.string().superRefine(...)])` with a single `z.string().superRefine(...)` that handles the empty string case internally
- Applied the same fix to both client-side (`AuthPageBrandingForm.tsx`) and server-side (`upsertLoginPageBranding.ts`) validation schemas

## Problem

When entering a logo URL that fails validation (e.g., the server doesn't return an `image/*` content-type), users see the cryptic error:

> `Too big: expected string to have <=0 characters`

This happens because Zod's `z.union` tries all branches when one fails. The first branch (`z.literal("")`) expects an empty string, so its error message about `<=0 characters` gets surfaced instead of the actual validation error from the `superRefine` branch.

## Fix

Move the empty string check inside the `superRefine` callback as an early return. This eliminates the `z.union` entirely, so validation errors (invalid URL, non-image content-type, network errors, etc.) are always reported directly to the user.

## How to test?

All existing validation cases are preserved:
- Empty string (no logo) -> passes
- Invalid URL -> `"Must be a valid URL"`
- URL returns non-200 -> `"Failed to load image..."`
- URL doesn't serve image content-type -> `"URL does not point to an image..."`
- Network/fetch error -> specific error message
- Enterprise local path validation -> unchanged

### DIFF
--- a/server/private/routers/loginPage/upsertLoginPageBranding.ts
+++ b/server/private/routers/loginPage/upsertLoginPageBranding.ts
@@ -35,76 +35,80 @@ const paramsSchema = z.strictObject({
 
 const bodySchema = z.strictObject({
     logoUrl: z
-        .union([
-            z.literal(""),
-            z
-                .string()
-                .superRefine(async (urlOrPath, ctx) => {
-                    const parseResult = z.url().safeParse(urlOrPath);
-                    if (!parseResult.success) {
-                        if (build !== "enterprise") {
-                            ctx.addIssue({
-                                code: "custom",
-                                message: "Must be a valid URL"
-                            });
-                            return;
-                        } else {
-                            try {
-                                validateLocalPath(urlOrPath);
-                            } catch (error) {
-                                ctx.addIssue({
-                                    code: "custom",
-                                    message: "Must be either a valid image URL or a valid pathname starting with `/` and not containing query parameters, `..` or `*`"
-                                });
-                            } finally {
-                                return;
-                            }
-                        }
-                    }
+        .string()
+        .superRefine(async (urlOrPath, ctx) => {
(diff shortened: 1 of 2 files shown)


LaurenceJJones:
Hey, just to let you know you seem to have a gpg signature applied to your github account but all commits over the last 3 PR's are not signed (or your gpg has expired), this is not great for security purposes.

If you used a llm to generate the changes ensure you go back and amend your signatures on the commits so they are at least signed, hence the feature of github.

also you didnt open the PR using the pull request template (another LLM sign) can you please include the license notice in your original pull request description, if not we will close as we need you to agree to the license before merge.

oschwartz10612:
I think we solved this one in 1.15.3 but thank you for the contribution!

PR #2523, opened 2026-02-23
fix: clean up ACME certs when resources are deleted

### LINKED ISSUE #2382: Pangolin requests certificates for domains of deleted resources
### Describe the Bug

When I delete a public resource from Pangolin, Pangolin still requests certificates for that subdomain. I regularly get certificate transparency reports for certificates issued from Lets Encrypt to subdomains that I used to have as a resource in pangolin and have since been deleted.

### Environment

- OS Type & Version: Ubuntu 24.04
- Pangolin Version: 1.15.1
- Gerbil Version: 1.3.0
- Traefik Version: 3.6.7
- Newt Version: 1.9.0

### To Reproduce

1. Create a public resource in Pangolin
2. After a certificate has been issued for the resource's subdomain and is available in Pangolin, delete the resource 
3. When certificate renewal comes around, Pangolin will renew the certificate of the deleted resource's subdomain

### Expected Behavior

Certificate issuance and renewals are only requested for subdomains connected to existing resources.

shreyaspapi:

## Description

Fixes #2382

## What's wrong

When you delete a resource, Traefik keeps its certificate in `acme.json` and keeps renewing it forever. The `deleteResource` handler only removes the DB row — it never touches the ACME storage.

## What this does

- After deleting a resource, checks if any other resource still uses the same domain
- If not, removes that domain's certificate from `acme.json` so Traefik stops renewing it
- Adds a configurable `acme_json_path` option (defaults to `/app/config/letsencrypt/acme.json`)
- The cleanup is best-effort — if it fails for any reason, the delete still succeeds

## How to test?

1. Create a resource with a custom domain that gets a Let's Encrypt cert
2. Delete that resource
3. Verify the domain's cert is removed from `acme.json`
4. Verify that domains shared by other resources are NOT removed

### DIFF
--- a/server/lib/traefik/acmeCleanup.test.ts
+++ b/server/lib/traefik/acmeCleanup.test.ts
@@ -2,62 +2,7 @@ import * as fs from "fs";
 import * as path from "path";
 import * as os from "os";
 import { assertEquals } from "@test/assert";
-
-/**
- * Standalone test for ACME JSON cleanup logic.
- *
- * Since the actual removeDomainFromAcmeJson function depends on the config
- * module (which requires DB setup), we test the core logic directly here
- * by extracting and testing the same algorithm.
- */
-
-// ---- Core logic extracted from acmeCleanup.ts for testability ----
-
-interface AcmeCertificate {
-    domain: {
-        main: string;
-        sans?: string[];
-    };
-    certificate: string;
-    key: string;
-}
-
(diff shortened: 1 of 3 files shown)


oschwartz10612:
Thanks for the PR. Right now I dont think we want pangolin to be messing with the acme file. We may solve this another way in the future.

PR #2524, opened 2026-02-23
fix: path-based routing broken due to key collisions in sanitize()

### LINKED ISSUE #2294: Path-Based-Routing is broken
### Describe the Bug

Multiple targets on a single domain act as load balancers regardless of match-path rules or priority.

### Environment

- OS Type & Version: (e.g., Ubuntu 22.04)
Ubuntu 22.04
- Pangolin Version:
1.14.1
- Gerbil Version:
Latest
- Traefik Version:
Latest
- Newt Version:
1.8.1
N/A

### To Reproduce

Set a path based rule, and try to trigger it. For example set the following rules on any resource:

Prefix: "/" -> 192.168.1.7:3000

Prefix: "/api" -> 192.168.1.7:3001

The backend service that is returned will be random, completely ignoring PBR rules.

I have also tried only setting a single prefix rule for "/api", and changing priorities to be equal, above, below, equal non-100, and non-100 above/below.

Regardless, the service returned is random.

### Expected Behavior

domain.com would route to the service on 192.168.1.7:3000 and domain.com/api (or /api/share etc.) would route to the service on 192.168.1.7:3001.

shreyaspapi:

## Description

Fixes #2294

## What this fixes

The `sanitize()` function used to generate internal map keys for grouping targets in `getTraefikConfig` can cause path collisions. It replaces all non-alphanumeric characters (including `/`, `.`, `_`) with dashes and collapses them, so structurally different paths can produce the same key:

- `sanitize("/a/b")` = `"a-b"`
- `sanitize("/a-b")` = `"a-b"` (same!)

When two targets get the same key, they're grouped into one loadbalancer service and Traefik round-robins between them instead of routing by path.

Note: this specific collision doesn't affect the `/` vs `/api` scenario reported in the issue (those already produce different keys), but it would affect setups with paths like `/api/v1` vs `/api-v1` or `/app.v2` vs `/app/v2`.

## How to test?

1. Create two resources with paths that would collide under old sanitization (e.g., `/api/v1` and `/api-v1`)
2. Verify they route to their correct backends independently (no round-robin)
3. Verify existing resources with simple paths still work with the same router names


shreyaspapi:
Fixed need a review

oschwartz10612 (approved):

oschwartz10612:
Thank you!

Issue #2528, opened 2026-02-24
UI toggle for "Enable Rules" and "Email Whitelist" does not reflect DB state on initial load


Kapot:
### Describe the Bug

When `applyRules` and/or `emailWhitelistEnabled` are set to `1` directly in the SQLite database, the corresponding toggles in the UI ("Enable Rules" on the Rules tab, "Email Whitelist" on the Authentication tab) render as OFF after a full `docker compose restart`. The data itself (rules and emails) is visible once the toggle is manually clicked on, confirming the DB values are correct and being read — only the initial toggle state is wrong.

### Environment

- OS Type & Version: Ubuntu 24.04.4 LTS
- Pangolin Version: EE 1.15.2
- Gerbil Version: 1.3.0
- Traefik Version: v3.6

### To Reproduce

1. Create a new resource
2. Insert rules into resourceRules for that resourceId via sqlite3
3. Insert emails into resourceWhitelist for that resourceId via sqlite3
4. Set applyRules = 1 and emailWhitelistEnabled = 1 on the resource via sqlite3
5. Verify DB values are correct: SELECT resourceId, name, applyRules, 
   emailWhitelistEnabled FROM resources WHERE resourceId = X;
6. Restart Pangolin: docker compose restart
7. Open the resource in the UI, navigate to the Rules tab — "Enable Rules" toggle 
   appears OFF despite applyRules = 1 in DB. Navigate to the Authentication tab — 
   "Email Whitelist" toggle appears OFF despite emailWhitelistEnabled = 1 in DB.
8. Click both toggles ON — rules and emails appear immediately, confirming data 
   was present all along.

### Expected Behavior

The UI should reflect the actual DB state on load. If applyRules = 1 and emailWhitelistEnabled = 1, the corresponding toggles should render as ON after restart without requiring manual interaction.

PR #2537, opened 2026-02-25
fix: sync resource toggle states with context on initial load

Abhinav-kodes:

## Description
Fixes a hydration bug where the **Enable Rules**, **SSO**, and **Email Whitelist** toggles on resource settings pages would always render in the OFF state on cold load, regardless of the actual saved value in the database.

Closes fosrl/pangolin#2528

## Root Cause

The `SwitchInput` components were using the `defaultChecked` prop (uncontrolled), which React only reads on the **first render**. Since the resource context hydrates asynchronously after mount, the first render always sees `undefined` or a stale value, so the toggles always appeared OFF.

Additionally, the state variables were initialized once via `useState(resource.X)` with no mechanism to re-sync when the context updated.

### `src/app/[orgId]/settings/resources/proxy/[niceId]/rules/page.tsx`
- `useState(resource.applyRules ?? false)` — added `?? false` fallback
- Added `useEffect` to sync `rulesEnabled` when `resource.applyRules` updates
- Changed `defaultChecked={rulesEnabled}` → `checked={rulesEnabled}` (controlled)

### `src/app/[orgId]/settings/resources/proxy/[niceId]/authentication/page.tsx`
- `useState(resource.sso ?? false)` — added `?? false` fallback
- Added `useEffect` to sync `ssoEnabled` when `resource.sso` updates
- Changed `defaultChecked={resource.sso}` → `checked={ssoEnabled}` (controlled)
- `useState(resource.emailWhitelistEnabled ?? false)` — added `?? false` fallback
- Added `useEffect` to sync `whitelistEnabled` when `resource.emailWhitelistEnabled` updates
- Changed `defaultChecked={resource.emailWhitelistEnabled}` → `checked={whitelistEnabled}` (controlled)

## Before / After

**Before:** Toggle always renders OFF on hard refresh, even when DB value is `true`.  
**After:** Toggle correctly reflects the saved DB value immediately on load.

## Testing

- [ ] Hard refresh (`Ctrl+Shift+R`) on a resource with `applyRules = true` → Rules toggle shows ON
- [ ] Hard refresh on a resource with `emailWhitelistEnabled = true` → Whitelist toggle shows ON
- [ ] Hard refresh on a resource with `sso = true` → SSO toggle shows ON
- [ ] Toggling any switch and saving still works correctly
- [ ] No React controlled/uncontrolled warnings in console

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/src/app/[orgId]/settings/resources/proxy/[niceId]/authentication/page.tsx
+++ b/src/app/[orgId]/settings/resources/proxy/[niceId]/authentication/page.tsx
@@ -187,7 +187,11 @@ export default function ResourceAuthenticationPage() {
         number | null
     >(null);
 
-    const [ssoEnabled, setSsoEnabled] = useState(resource.sso);
+    const [ssoEnabled, setSsoEnabled] = useState(resource.sso ?? false);
+
+    useEffect(() => {
+        setSsoEnabled(resource.sso ?? false);
+    }, [resource.sso]);
 
     const [selectedIdpId, setSelectedIdpId] = useState<number | null>(
         resource.skipToIdpId || null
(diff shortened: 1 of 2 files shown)


oschwartz10612:
Thanks!

PR #2552, opened 2026-02-26
feat: Adding an organization sites bandwidth reset API

huzky-v:

## Description
This feature change adds an API to manually reset bandwidth for all sites in the selected organization so that the users can reset the bandwidth counter when needed.
It is useful for the user hosted on VPS to see the monthly bandwidth counter instead of seeing accumulated counter by resetting the counter monthly. 

The API call is `POST /org/{orgId}/reset-bandwidth`, OpenAPI and permission is added to the system.


oschwartz10612:
LGTM thanks so much!

PR #2553, opened 2026-02-26
enhance(sidebar): make mobile org selector sticky

LaurenceJJones:

## Description

Make org selector sticky on mobile sidebar

Move OrgSelector outside the scrollable container so it stays fixed at the top while menu items scroll, matching the desktop sidebar behavior introduced in 9b2c0d0b.

this may have been deliberately chosen to not sticky on mobile when the commit above was done, if so feel free to close.

## How to test?

When viewing the mobile dashboard, the organization selector will be sticky at the top so even if the user scrolls the menu they have easier access to moving organizations.


oschwartz10612:
@miloschwartz?

PR #2568, opened 2026-03-01
feat: OIDC provider for "Login with Pangolin"

mallendeo:

## Description

Implements Pangolin as an OIDC provider for “Login with Pangolin”, with full OAuth client management for admins and app-consent management for users.

See https://github.com/orgs/fosrl/discussions/21

Scaffolding and ongoing development were AI-assisted (Opus 4.6 / Codex 5.3); not all code has been manually reviewed. Contributions welcome.

## What’s included

- OAuth 2.0 Authorization Code flow with PKCE
- OIDC endpoints: discovery, authorize, token, userinfo, JWKS, revoke
- OIDC logout support:
  - RP-initiated logout (`end_session_endpoint`)
  - back-channel logout dispatch
- Admin OAuth Clients UI: create, edit, delete, rotate secret, advanced settings
- User "Connected Apps" page to revoke grants
- Edit profile dialog (name, username) to support profile claims
- Claims updates, including `given_name`, `family_name`, and org-scoped `groups`
- i18n strings (English only, AI translation removed / out of scope)
- SQLite + PostgreSQL schema/migration updates for v1.17.0

## How to test

If upgrading from a previous install, route `/.well-known` to the API service.

In `config/traefik/dynamic_config.yml`, update the Next.js and API router rules:

```yaml
next-router:
  rule: "Host(`yourdomain.com`) && !PathPrefix(`/api/v1`) && !PathPrefix(`/.well-known`)"
  # ...

api-router:
  rule: "Host(`yourdomain.com`) && (PathPrefix(`/api/v1`) || PathPrefix(`/.well-known`))"
  # ...
```

Restart Traefik after changing this.

### Create an OAuth client

1. Go to Org Settings > OAuth Clients
2. Click Create Client
3. Fill in name, redirect URI, and scopes
4. Save the client ID and secret shown in the dialog

### Verify discovery

curl https://yourdomain.com/.well-known/openid-configuration

Expected: JSON with fields like `issuer`, `authorization_endpoint`,
`token_endpoint`, `userinfo_endpoint`, `jwks_uri`, `revocation_endpoint`,
and `end_session_endpoint`.

### Available scopes and claims

| Scope   | Claims                                                                      |
| ------- | --------------------------------------------------------------------------- |
| openid  | sub                                                                         |
| profile | name, preferred_username, given_name, family_name                           |
| email   | email, email_verified                                                       |
| groups  | groups (organization/role memberships, formatted like <orgName>:<roleName>) |

### Endpoints

| Endpoint      | Path                                |
| ------------- | ----------------------------------- |
| Discovery     | /.well-known/openid-configuration   |
| Authorization | /oauth/authorize (browser redirect) |
| Token         | /api/v1/oauth/token                 |
| Userinfo      | /api/v1/oauth/userinfo              |
| JWKS          | /api/v1/oauth/jwks                  |
| Revoke        | /api/v1/oauth/revoke                |
| Logout        | /api/v1/oauth/logout                |

## Demo

TODO


Root-Core:
Is this still a draft or ready for review? Anything on your todo?
The PR is a pretty big and the code bases are likely to diverge quickly.

mallendeo:
> Is this still a draft or ready for review? Anything on your todo? The PR is a pretty big and the code bases are likely to diverge quickly.

Still a draft, it's functional but I haven't had time to do a thorough review of the full diff myself, so I wouldn't ship it as-is.

If anyone wants to pick this up for review and cleanup, that would be welcome. I won't be able to dedicate time to this in the following weeks.

server/routers/oauth/revoke.ts
```diff
@@ -0,0 +1,137 @@
...
+                    .delete(oauthAccessTokens)
+                    .where(grantWhere(oauthAccessTokens, grantId));
+                await trx
+                    .update(oauthRefreshTokens)
+                    .set({ revokedAt: Date.now() })
+                    .where(grantWhere(oauthRefreshTokens, grantId));
+            });
+        }
+
+        if (tokenTypeHint === "refresh_token") {
```
Root-Core, line 61: This could be simplified, so that all database queries use a single function.

server/routers/oauth/revoke.ts
```diff
@@ -0,0 +1,137 @@
...
+                .from(oauthRefreshTokens)
+                .where(
+                    and(
+                        eq(oauthRefreshTokens.tokenHash, tokenHash),
+                        eq(oauthRefreshTokens.clientId, client.clientId)
+                    )
+                )
+                .limit(1);
+
+            if (refreshTokenRecord) {
```
Root-Core: There should be a `timingSafeEqual()` to mitigate timing based attack vectors.
Something like this, which also makes sense in a generic function:

```ts
    const isValid = timingSafeEqual(
        Buffer.from(tokenRecord.tokenHash),
        Buffer.from(tokenHash)
    );
```

server/routers/oauth/revoke.ts
```diff
@@ -0,0 +1,137 @@
...
+            .where(
+                and(
+                    eq(oauthAccessTokens.tokenHash, tokenHash),
+                    eq(oauthAccessTokens.clientId, client.clientId)
+                )
+            )
+            .limit(1);
+
+        if (accessTokenRecord) {
+            await revokeGrantById(accessTokenRecord.grantId);
```
Root-Core: See above

server/routers/oauth/revoke.ts
Root-Core: I first had the concern that there might be an attack surface via timing leak, but as the lookup happens at the database layer and the token is revoked if found.. it should be fine.

No need for a `timingSafeEqual()`. The reuse of code still applies though.

server/routers/oauth/token.ts
```diff
@@ -0,0 +1,376 @@
...
+            if (!code || !redirectUri) {
+                return sendOAuthError(res, HttpCode.BAD_REQUEST, {
+                    error: "invalid_request",
+                    error_description: "Missing code or redirect_uri"
+                });
+            }
+
+            // Atomically consume the auth code to prevent race conditions (TOCTOU).
+            // Per RFC 6749 Section 10.5, a code presented with invalid params
+            // after atomic deletion is correctly invalidated.
```
Root-Core, line 168: This is good as it _prevents_ replay attacks, deleting the code however prevents _detecting_ replay attacks. This isn't really an issue, but might be relevant at some point.

server/routers/oauth/token.ts
```diff
@@ -0,0 +1,376 @@
...
+                });
+            }
+
+            const now = Date.now();
+
+            // Atomically revoke the refresh token to prevent reuse race conditions.
+            // Same pattern as auth code consumption above (DELETE...RETURNING).
+            const [existingRefreshToken] = await db
+                .update(oauthRefreshTokens)
+                .set({ revokedAt: now })
```
Root-Core, line 275: Is there a reason to not delete the token directly like above?

- It also prevents reuse
- It doesn't need a cleanup
    - The cleanup will still query `expiredAt` though
- Less risk of timing based attacks

The same comment is true though, here we could _detect_ potential attacks.
Maybe decide to use one of the implementation patterns.

server/lib/oauth/backchannelLogout.ts
Root-Core, line 1: If there is an network issue, the user isn't logged out from the remote app.
There is no retry mechanism and the logout fails silently, so the user will only be logged out if the token expires (up to 30 days, I think). The user is also not informed about this issue.

This is could be a security concern.

Waiting for all responses would be really slow.
Queuing retries is a bit more complex and also doesn't guarantee success.
If the queue fails though, this is a signal that there is a bigger or persistent issue.

server/lib/oauth/keys.ts
```diff
@@ -0,0 +1,112 @@
...
+
+export async function ensureSigningKey(): Promise<void> {
+    const [existingKey] = await selectActiveSigningKeys();
+
+    if (existingKey) {
+        return;
+    }
+
+    const keyPair = generateKeyPairSync("rsa", {
+        modulusLength: 2048,
```
Root-Core, line 34: Why not `RSA-3072` or `RSA-4096`?
Some governments consider `RSA-2048` deprecated for new implementations.

server/lib/oauth/keys.ts
```diff
@@ -0,0 +1,112 @@
...
+        .select()
+        .from(oauthSigningKeys)
+        .where(eq(oauthSigningKeys.active, true))
+        .orderBy(desc(oauthSigningKeys.createdAt));
+}
+
+export async function ensureSigningKey(): Promise<void> {
+    const [existingKey] = await selectActiveSigningKeys();
+
+    if (existingKey) {
```
Root-Core, line 29: The database mechanisms imply that the keys could rotate, however there is only ever one key created and used. The generation should be moved to a separate function.

Root-Core: Also: not sure if the secret from `config.getRawConfig().server.secret` can ever change, but this would also break the signing key. Validation via the public key would still be possible.

server/lib/oauth/keys.ts
Root-Core, line 1: Is key rotation a planed feature? What would be the migration path?

The current implementation allows only the validation against the active signing key.
If the signing key rotates, existing tokens can't be validated.
This could be solved by using the keyId in the JWT to get the signing key, instead of the active one.

server/routers/oauth/endSession.ts
```diff
@@ -0,0 +1,191 @@
...
+                .limit(1);
+
+            if (client) {
+                clientLogoutTerminatesPangolinSession =
+                    client.logoutTerminatesPangolinSession;
+                const registeredUris = client.postLogoutRedirectUris ?? [];
+
+                if (
+                    postLogoutRedirectUri &&
+                    registeredUris.includes(postLogoutRedirectUri)
```
Root-Core, line 157: Is there a usecase, where we want to match against subdomains, wildcard or regex domains?

server/routers/oauth/clients.ts
```diff
@@ -0,0 +1,546 @@
...
+
+        await db.transaction(async (trx) => {
+            await trx
+                .update(oauthClients)
+                .set({
+                    clientName: body.clientName,
+                    clientUri: body.clientUri,
+                    logoUri: body.logoUri,
+                    redirectUris: body.redirectUris,
+                    scopes: body.scopes
```
Root-Core, line 337: If `scopes` is changed, the existing consents aren't updated.
This could allow clients to request information that is now prohibited.

I think the existing `scopes` in `oauthConsents` should be updated accordingly.

Root-Core: I'm not sure how to handle consents that have no valid scope left.
Delete the entry or keep it with `[]`?

Root-Core: To be more conservative, we could delete all affected consents altogether.
This is not the best UX, but arguably the most secure approach.

server/routers/oauth/authorize.ts
```diff
@@ -0,0 +1,549 @@
...
+        if (!isClientOrgMember) {
+            return next(
+                createHttpError(
+                    HttpCode.FORBIDDEN,
+                    "User does not belong to this OAuth client's organization"
+                )
+            );
+        }
+
+        if (!client.redirectUris.includes(body.redirect_uri)) {
```
Root-Core, line 183: Same as in `endSession.ts`: 

> Is there a usecase, where we want to match against subdomains, wildcard or regex domains?

src/lib/cleanRedirect.ts
```diff
@@ -14,6 +14,10 @@ const ALLOWED_QUERY_PARAMS = new Set([
 
 const DUMMY_BASE = "https://internal.local";
 
+export function cleanOAuthRedirectOptions(path: string | undefined | null): CleanRedirectOptions {
+    return path?.startsWith("/oauth/authorize") ? { allowAllQueryParams: true } : {};
```
Root-Core, line 18: Should the params be filtered or is `allowAllQueryParams` okay?

src/lib/cleanRedirect.ts
```diff
@@ -14,6 +14,10 @@ const ALLOWED_QUERY_PARAMS = new Set([
 
 const DUMMY_BASE = "https://internal.local";
 
+export function cleanOAuthRedirectOptions(path: string | undefined | null): CleanRedirectOptions {
```
Root-Core, line 17: Why not only `path: string`?

server/lib/oauth/backchannelLogoutSecurity.ts
Root-Core, line 1: Node.js has the class `net.BlockList` - wouldn't it be easier to use that?

server/routers/external.ts
Root-Core, line 1: Consider implementing an (additional) rate limit per OAuth client (app), as one malicious app could block access for all apps. Only using the IP might lead to many false positives - apps deployed on the same server.

Root-Core (commented):
I'm no security or OIDC expert - so ignore if I got something blatantly wrong.

I have not tested the implementation, just looked at the code.
I also focused on the backend, which is the important part for security.

I also havn't looked at the logic or implementation of the OAuth / OIDC spec, just a review of the code.

-----

I think the most important findings are

- Missing consent invalidation / validation
- Silent back channel logout failures, without retries
- Missing per client rate limit (instead of IP only)
- Discussion about potential signing key rotation

Nothing critical.

AstralDestiny:
https://openid.net/specs/openid-connect-core-1_0.html#IDToken I'll look later when I wake but is this following the openid spec or just to get it to work as a provider but not be openid compliant? (I will preface I haven't looked been busy and tired)

oschwartz10612:
We are looking at #3566. Thanks for all of the hard work on this! :}

PR #2570, opened 2026-03-01
fix: local targets ignored when newt site is unhealthy (mixed target failover)

### LINKED ISSUE #2448: Mixed Target Failover Broken (Local target ignored when Newt target is Unhealthy)
### Describe the Bug

I am running a mixed configuration for a single resource with two targets:
- Newt Target: has health check.
- Local Target, no health check possible (Status: Unknown / -).
When the Newt target goes down and reports as Unhealthy, Pangolin does not use the Local target. 
Instead, it returns: no available server.
**I have to disable the unhealthy target to switch to local target.**

### Environment

- OS Type & Version:  Ubuntu 22.04
- Pangolin Version: v1.15.2
- Gerbil Version: latest
- Traefik Version: 3.6.7
- Newt Version: 1.9.0

### To Reproduce

1. Create a resource with two targets:
    Target A: Site = Newt, (Status: Healthy)
    Target B: Site = Local, (Status: Unknown)
2. Verify traffic flows to Target A (Newt).
3. Simulate Failure for Target A to become Unhealthy

Result: Request returns: no available server.. Traffic is not routed to Target B, even though is reachable from the Pangolin host.

### Expected Behavior

Use local site if all other sites report unhealthy

Fizza-Mukhtar:

Fixes #2448

## Problem
Local targets were ignored when Newt targets became unhealthy because 
local sites always have online=false and were excluded by the anySitesOnline check.

## Fix
Added early return for local site type in the target filter so they are 
always considered available as fallback servers.

### DIFF
--- a/server/lib/traefik/getTraefikConfig.ts
+++ b/server/lib/traefik/getTraefikConfig.ts
@@ -477,7 +477,10 @@ export async function getTraefikConfig(
 
                         // TODO: HOW TO HANDLE ^^^^^^ BETTER
                         const anySitesOnline = targets.some(
-                            (target) => target.site.online
+                            (target) =>
+                            target.site.online ||
+                            target.site.type === "local" ||
+                            target.site.type === "wireguard"
                         );
 
                         return (
@@ -491,10 +494,6 @@ export async function getTraefikConfig(
                                         return false;
                                     }
                                     
-                                    // Local sites don't report online status, always include them as fallback
-                                    if (target.site.type === "local") {
-                                        return true;
-                                    }
                                     // If any sites are online, exclude offline sites
                                     if (anySitesOnline && !target.site.online) {
                                         return false;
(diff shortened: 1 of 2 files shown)


LaurenceJJones:
Looking at the diff, I wouldn't return true just because the target is local. There's already an if case a few lines below that checks for both local and wireguard sites with a satisfactory target, returning false if that condition isn't met.

I believe the better fix would be to classify any local target as online... maybe? 🤷🏻

Additionally, if we proceed with early return we should extend your if case to also cover satisfactory targets and wireguard sites. Your changes focus on ensuring local sites can handle failover, but wireguard has the same issue with lack of online state. 

edit: may be mistaken on the wireguard parts will dive into code to double check.

Fizza-Mukhtar:
Hey @LaurenceJJones Thanks for the feedback! Updated the fix to classify local and wireguard sites as online in the anySitesOnline check instead of using early return. Let me know if wireguard needs different handling after your review!

oschwartz10612:
I think this makes sense. Thanks for the fix! :} Good note @LaurenceJJones

PR #2573, opened 2026-03-01
fix: exclude labels from container search to prevent false positives

### LINKED ISSUE #2228: Bad searching experience when adding Targets
### Describe the Bug

I am not sure if this must be reported as feature - let me know if I need to raise this as FR.

Current implementation of search is quite useless when searching for containers if they are created with Docker Compose (which should be a common way). So I am using a per service compose file with application name as a filename. Let's say - containers forgejo with forgejo-db will be stored in forgejo.yml. When compose stack is created, every container in the stack will have label `com.docker.compose.project.config_files` with a list of files. I have near 60 containers and it means that all of them will have `com.docker.compose.project.config_files` label with `forgejo.yml` value (path to actual file) in it (as well as the rest of the files).
```
"com.docker.compose.project.config_files": ".../home/user/docker/compose/stack/services/forgejo.yml...",
```
So when searching for forgejo in Containers list when adding Targets, all 60 containers are returned (as search apparently includes labels) which makes searching functionality useless and Ctrl+F is the only way to find the container in the long list. It might be better to have some option to control which column/fields to search for or isolate label search into a separate section.

### Environment

- OS Type & Version: Almalinux 10
- Pangolin Version: 1.4.1

### To Reproduce

Make a small compose stack with several files per service and search for container with the name (assuming yml file will have the same name as container name which quite common).

### Expected Behavior

The current behavior looks expected, however user experience is bad.

Fizza-Mukhtar:

Fixes #2228

## Problem
Container search was including Docker labels in searchable fields. This caused false positives when using Docker Compose, since all containers in a stack share the same compose file path in their labels.

## Fix
Removed labels from the search fields. Labels are already accessible via their own dedicated column and popover, so excluding them from search gives more accurate results.

### DIFF
--- a/src/components/ContainersSelector.tsx
+++ b/src/components/ContainersSelector.tsx
@@ -171,8 +171,7 @@ const DockerContainersTable: FC<{
                 ...Object.values(container.networks)
                     .map((n) => n.ipAddress)
                     .filter(Boolean),
-                ...getExposedPorts(container).map((p) => p.toString()),
-                ...Object.entries(container.labels).flat()
+                ...getExposedPorts(container).map((p) => p.toString())
             ];
 
             return searchableFields.some((field) =>


oschwartz10612:
Thank you!

PR #2574, opened 2026-03-01
feat: add pageSize persistence to ControlledDataTable

LaurenceJJones:

## Description

  - ControlledDataTable now persists pageSize to localStorage (default: true)
  - Refactored DataTable to use shared storage hooks
  - Exported getStoredPageSize/setStoredPageSize from useStoredPageSize
  - Exported getStoredColumnVisibility/setStoredColumnVisibility from useStoredColumnVisibility
  - Unified storage key format: datatable-{tableId}-page-size, datatable-{tableId}-column-visibility

## How to test?

Page sizes or column visibility across data-table and controlled-data-table are now using the same hooks for persistence to local storage, this ensures storage naming schemes across the tables stay consistent, instead of each component owning the process. When altering the page size or column visibility on any data table this should be stored within local storage to ensure navigating away or coming back stays consistent.

The only issue we will have that previous data-table local storage items will be stale and not used since the newer scheme was `datatable-${tableid}` prefix, let me know if we should handle an automatic migration?

### DIFF
--- a/src/components/ui/controlled-data-table.tsx
+++ b/src/components/ui/controlled-data-table.tsx
@@ -37,7 +37,7 @@ import {
 } from "@app/hooks/useStoredPageSize";
 import { Columns, Filter, Plus, RefreshCw, Search } from "lucide-react";
 import { useTranslations } from "next-intl";
-import { useEffect, useRef, useState } from "react";
+import { useEffect, useMemo, useRef, useState } from "react";
 
 // Extended ColumnDef type that includes optional friendlyName for column visibility dropdown
 export type ExtendedColumnDef<TData, TValue = unknown> = ColumnDef<


oschwartz10612:
@miloschwartz?

miloschwartz:
Thanks @LaurenceJJones! Looks like this always overrides the query param, so we can't set a custom page size of 25 for example. Could we tweak this so it prefers the value from the query param first, then falls back to the saved value if there is one? 

`?page=1&pageSize=25`

Fredkiss3:
Hello there 👋🏾 

The pageSize and pagination state is passed from the parent, it's not persisted by `ControlledDataTable` because the value of pageSize is not controlled by the data table, it's fully controlled by the server.  

The data-table doesn't really control any state on the table and all of its state comes from the server/url search params (that's why it's called controlled data table). 

In order to persist this state, we would need to persist it in something like cookies for ex so that the server can be aware of that.

miloschwartz:
Good point @Fredkiss3

LaurenceJJones:
> Hello there 👋🏾
> 
> The pageSize and pagination state is passed from the parent, it's not persisted by `ControlledDataTable` because the value of pageSize is not controlled by the data table, it's fully controlled by the server.
> 
> The data-table doesn't really control any state on the table and all of its state comes from the server/url search params (that's why it's called controlled data table).
> 
> In order to persist this state, we would need to persist it in something like cookies for ex so that the server can be aware of that.

Yes, I realized after creating the PR that a double load was happening and that because client loads -> local storage fires -> client loads again with query parameter. So I totally agree this is not the right direction.

Issue #2551, opened 2026-02-26
Bypass Auth path rules do not work for unauthenticated users on Protected resources


BeBeRex56:
### Describe the Bug

When a resource is set to Protected, path-based Bypass Auth rules do not allow unauthenticated users to access those paths. Unauthenticated visitors are always redirected to the Pangolin authentication wall regardless of matching Bypass Auth rules. Authenticated users with an existing session are unaffected, which makes the issue easy to miss during testing.

### Environment

- OS Type & Version: Ugreen NAS (Docker)
- Pangolin Version: latest (hosted via app.pangolin.net)
- Gerbil Version: N/A
- Traefik Version: N/A
- Newt Version: 1.10.0
- Pangolin-CLI Version: latest
- Olm Version: N/A

### To Reproduce

1. Create a resource with Authentication set to Protected
2. Navigate to the Rules tab for the resource
3. Add a rule: Action = Bypass Auth, Match Type = Path, Value = /share/
4. Save the rules
5. Open a path matching the rule (e.g. /share/<token>) in a private/incognito browser window
6. Observe that Pangolin redirects to the authentication page instead of allowing access

### Expected Behavior

A Bypass Auth path rule should allow unauthenticated users to access the matched path without being redirected to the login page, even when the resource is globally Protected. This is needed to support use cases where a protected resource has specific public-facing endpoints, such as application-generated share links.

<html>
<body>
<html><head></head><body><h1>Bypass Auth and Pass to Auth Path Rules Do Not Work Correctly in Either Authentication Mode</h1>
<h2>Summary</h2>
<p>Path-based rules in Pangolin's Resource Rules Configuration are broken in both directions:</p>
<ol>
<li>When a resource is set to <strong>Protected</strong>, <code>Bypass Auth</code> rules do not allow unauthenticated users to access matched paths — the global auth wall fires before rules are evaluated.</li>
<li>When a resource Authentication is set to <strong>off/none</strong>, <code>Pass to Auth</code> rules do not enforce authentication on matched paths — everything is publicly accessible regardless of rules.</li>
</ol>
<p>This makes it impossible to achieve a mixed-access configuration where some paths on a resource are public (e.g. share links) and others are protected — regardless of which direction you configure it from.</p>
<h2>Expected Behavior</h2>
<ul>
<li>When Authentication is <strong>Protected</strong>: <code>Bypass Auth</code> path rules should allow unauthenticated users to access matched paths without hitting the auth wall.</li>
[...]

PR #2578, opened 2026-03-02
fix: use web api to extract path from URL properly

moohbr:

## Description

Fix PATH based resource rules failing to match when the request contains a query string.

Badger sends the path field with the full URI including query parameters (e.g. `/stream.html?token=111`). The `isPathAllowed` function performs segment-based matching, so `/stream.html?token=111` never matches a rule configured as `/stream.html`, causing `Bypass Auth` rules to silently fail for unauthenticated users.

Instead of relying on the path and host fields sent by Badger, both are now derived from `originalRequestURL` using the native `URL API`, which guarantees a clean `pathname` (no query string or fragment) and a consistent host.

## How to test?

1. Create a resource with `Authentication` = Protected and Rules enabled
2. Add a rule: Bypass Auth → Path → `/stream.html`
3. Send a request to the internal verify-session endpoint with path: `/stream.html?token=111` and `originalRequestURL`: `https://<resource-domain>/stream.html?token=111` and empty sessions
4. Confirm the response is `valid: true`
5. Send the same request with path: `/config.html` and confirm `valid: false`

### DIFF
--- a/server/routers/badger/verifySession.ts
+++ b/server/routers/badger/verifySession.ts
@@ -96,14 +96,15 @@ export async function verifyResourceSession(
     try {
         const {
             sessions,
-            host,
             originalRequestURL,
             requestIp,
-            path,
             headers,
             query,
             badgerVersion
         } = parsedBody.data;
+
+
+        const {pathname: path, host} = new URL(originalRequestURL);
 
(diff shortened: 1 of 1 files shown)


moohbr:
#2551

PR #2603, opened 2026-03-05
fix: prevent resource from being created with dashboard's domain to avoid redirect loop

### LINKED ISSUE #2595: Resource with dashboard's subdomain causes endless redirect loop
### Describe the Bug

The dashboard can be accessed at `pangolin.example.com`.
If you configure the same subdomain as a resource, you will be redirected in a loop (HTTP 302).

I configured it this way on purpose to test whether anything would break, and it did. :)
I don't think that's a common issue.

I fixed it by changing the dashboard's subdomain in the configuration; you can then remove the resource.

### Environment

- Pangolin Version: 1.16.2
- Gerbil Version: 1.3.0
- Traefik Version: 3.6

### To Reproduce

1. Setup Pangolin on `pangolin.example.com`
2. Add local site
3. Create public resource
    - HTTPS Settings
        - Subdomain `pangolin`
        - Base Domain `example.com`
    - Targets Configuration
        - Local site
        - Protocol `https`
        - Host `localhost`
        - Port `443`

### Expected Behavior

There should be a warning or error message.

Fizza-Mukhtar:

Fixes #2595

## Problem
Creating or updating a resource with the same subdomain as the dashboard caused an endless HTTP 302 redirect loop.

## Fix
- Added validation in `createResource` and `updateResource` to return a conflict error if the resource's full domain matches the dashboard URL
- Updated the create resource page to show the actual backend error message instead of a generic one

### DIFF
--- a/server/routers/resource/createResource.ts
+++ b/server/routers/resource/createResource.ts
@@ -224,7 +224,7 @@ async function createHttpResource(
     }
 
     // Prevent creating resource with same domain as dashboard
-    const dashboardUrl = process.env.DASHBOARD_URL;
+    const dashboardUrl = config.getRawConfig().app.dashboard_url;
     if (dashboardUrl) {
         const dashboardHost = new URL(dashboardUrl).hostname;
         if (fullDomain === dashboardHost) {
--- a/server/routers/resource/updateResource.ts
+++ b/server/routers/resource/updateResource.ts
@@ -354,7 +354,7 @@ async function updateHttpResource(
             }
 
             // Prevent updating resource with same domain as dashboard
-            const dashboardUrl = process.env.DASHBOARD_URL;
+            const dashboardUrl = config.getRawConfig().app.dashboard_url;
             if (dashboardUrl) {
                 const dashboardHost = new URL(dashboardUrl).hostname;
                 if (fullDomain === dashboardHost) {


LaurenceJJones:
It might be the same value but in most places instead of calling `process.env.DASHBOARD_URL` we simply do the following:

```js
import config from "@server/lib/config";
....
const dashboardUrl = config.getRawConfig().app.dashboard_url;
if (dashboardUrl) {
    const dashboardHost = new URL(dashboardUrl).hostname;
    if (fullDomain === dashboardHost) {
        return next(
            createHttpError(
                HttpCode.CONFLICT,
               "Resource domain cannot be the same as the dashboard domain"
           )
       );
    }
}
```

but maybe owen or milo can confirm if we should align to rest of codebase.

Fizza-Mukhtar:
> It might be the same value but in most places instead of calling `process.env.DASHBOARD_URL` we simply do the following:
> 
> ```js
> import config from "@server/lib/config";
> ....
> const dashboardUrl = config.getRawConfig().app.dashboard_url;
> if (dashboardUrl) {
>     const dashboardHost = new URL(dashboardUrl).hostname;
>     if (fullDomain === dashboardHost) {
>         return next(
>             createHttpError(
>                 HttpCode.CONFLICT,
>                "Resource domain cannot be the same as the dashboard domain"
>            )
>        );
>     }
> }
> ```
> 
> but maybe owen or milo can confirm if we should align to rest of codebase.

Good catch! Updated to use config.getRawConfig().app.dashboard_url to align with the rest of the codebase.

oschwartz10612 (approved):

oschwartz10612:
Thank you!

PR #2616, opened 2026-03-07
  fix(newt): missing hcStatus in hc config on reconnect

LaurenceJJones:

## Description

  The buildTargetConfigurationForNewtClient function was not including the
  hcStatus field when building health check targets for the newt/wg/connect
  message. This caused custom expected response codes (e.g., 409) to revert
  to the default 2xx range check after Pangolin server restart.

 Added hcStatus to both the database select query and the returned health
 check target object, matching the behavior in targets.ts addTargets.

fixes: https://github.com/fosrl/newt/issues/240

## How to test?

Follow the steps defined in https://github.com/fosrl/newt/issues/240, after this change the expected code should function correctly.


oschwartz10612:
Thanks!

PR #2618, opened 2026-03-08
Skip invalid Docker resources instead of failing entire blueprint (#1784)

### LINKED ISSUE #1784: A single Docker label error prevents all from updating
### Describe the Bug

I've been using the Docker socket functionality and it's really cool! However, it seems that if ANY Docker container has a label that Pangolin doesn't like, it prevents all other Docker services from being parsed by Pangolin.

I was updating a port number on an unrelated service (`service-2`), and I was puzzled why the port number wouldn't update in Pangolin even though I changed the label and brought the service up again. After a while I found this in the Pangolin log:

```
2025-10-30T18:47:36+00:00 [error]: Failed to update database from config: Error: Validation error: Required at "proxy-resources.service-2.targets[0].port"
```

The problem seemed to be that the `port` label NEEDED to be present on `service-2` (I'm guessing because the container didn't expose one that it could pick up automatically?), but this made `service-1` unable to change. I added the `port` label to `service-2` and that made `service-1` update, now that the Pangolin log was error free.

### Environment

- Pangolin Version: v1.11.1
- Gerbil Version: 
- Traefik Version: v3.3.3
- Newt Version:

### To Reproduce

See description

### Expected Behavior

Unrelated label configuration errors shouldn't affect other services.

LunarECL:

## Description

Fixes #1784

If one Docker container had bad labels (like missing a port on a container with no EXPOSE), it would block every other container from updating too. Not great.

Now there's a per-resource pre-validation step in `applyNewtDockerBlueprint.ts` that catches invalid resources and filters them out before they hit `applyBlueprint()`. Valid ones still go through the normal `ConfigSchema` validation, so cross-resource checks like domain/port/alias uniqueness work the same as before. Invalid ones just get logged with the Zod error path and skipped.

Only touched `server/lib/blueprints/applyNewtDockerBlueprint.ts`. Everything else (`applyBlueprint.ts`, `types.ts`, `parseDockerContainers.ts`) is unchanged.

## How to test?

1. Set up two Docker services with Pangolin labels on the same Docker socket
2. Remove the `port` label from one service that has no `EXPOSE` directive
3. Update a label on the other (valid) service
4. **Before fix:** Both services fail to update. Log shows `Validation error: Required at "proxy-resources.testservice.targets[0].port"`
5. **After fix:** Valid service updates normally. Invalid service gets skipped with a warning: `Skipping invalid Docker proxy-resources "testservice": targets.0.port: Required`Required`

### DIFF
--- a/server/lib/blueprints/applyNewtDockerBlueprint.ts
+++ b/server/lib/blueprints/applyNewtDockerBlueprint.ts
@@ -1,9 +1,55 @@
 import { sendToClient } from "#dynamic/routers/ws";
 import { processContainerLabels } from "./parseDockerContainers";
 import { applyBlueprint } from "./applyBlueprint";
+import { ResourceSchema, ClientResourceSchema } from "./types";
 import { db, sites } from "@server/db";
 import { eq } from "drizzle-orm";
 import logger from "@server/logger";
+
+type BlueprintResult = ReturnType<typeof processContainerLabels>;
+
+function filterInvalidResources(blueprint: BlueprintResult): {
+    skippedCount: number;
+    skippedKeys: string[];
+} {
+    const skippedKeys: string[] = [];
+
+    for (const section of ["proxy-resources", "public-resources"] as const) {
+        const resources = blueprint[section];
+        for (const [key, value] of Object.entries(resources)) {
+            const result = ResourceSchema.safeParse(value);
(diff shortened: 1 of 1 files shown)


oschwartz10612 (approved):

oschwartz10612:
Thank you!

PR #2623, opened 2026-03-09
refactor(install): improve resource cleanup and remove unused funcs

rodneyosodo:

## Description
- Remove unused functions
- Improve resource cleanup

## How to test?
N/A

### DIFF
--- a/install/config.go
+++ b/install/config.go
@@ -99,11 +99,6 @@ func ReadAppConfig(configPath string) (*AppConfigValues, error) {
 	return values, nil
 }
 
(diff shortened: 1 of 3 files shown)


LaurenceJJones:
We dont generally use named returns across our code bases, cleaning up unused functions can be useful but when the binary is built go will strip these anyways.

rodneyosodo:
We rely on named returns to capture errors from deferred calls that might otherwise be unchecked. This avoids the need to manually invoke the function at the end of the scope.

oschwartz10612:
Thanks!

PR #2653, opened 2026-03-15
feat(installer): Update docker-compose.yml with HTTPS/3 + QUIC support via traefik

shleeable:
## Context
Based on https://github.com/orgs/fosrl/discussions/363 - please test and verify this to build some confidence for the Pangolin team.

`Hey, there is not specific reason this isn't implemented other than we haven't tested it and had time to add it. I don't see why would can't in the future! Thanks for the suggestion :)` - @miloschwartz 

but the good news is it's been stable on traefik's side for many releases now. So I doubt there will be a regression, if there is, we should speak with the traefik team.

## Description
* Updates the port forwarding for the 443 UDP port to traefik
* Enabled the http3 traefik entrypoint

## How to test?
* Cherrypick PR
* Restart the docker compose services
* Open your firewall ports (if required)
* Test your endpoints using tools. eg: https://http3check.net/?host=loops.plus
* Celebrate the easy win

image


LaurenceJJones:
Will review shortly, the only caveat at the moment is there is two traefik configs one if crowdsec is selected and without.

Might be best if we consolodate these into a single dynamic template, to ensure we don't miss either with wins like this.

oschwartz10612 (approved):
LGTM. I think @LaurenceJJones is right but we can tackle that in the future.

oschwartz10612:
Thanks!

PR #2676, opened 2026-03-19
feat: header token authentication to bypass SSO for API access

crisidev:

## Description
Adds a new per-resource authentication method that allows clients to bypass SSO by sending a secret token in a custom HTTP header. This is designed for programmatic access: CI/CD pipelines, monitoring tools, webhooks, and other integrations that can't go through browser-based auth.                                                               
                                                                                
  - Resources can opt-in by setting a header name (e.g. `X-Pangolin-Token`) in the authentication settings             
  - Tokens are generated with 128-bit entropy, stored as SHA-256 hashes, with optional expiration
  - Badger's `verifySession` checks the header before falling through to SSO, with a 5-second local cache keyed by hash
   (not raw token)                                                                                                     
  - Full CRUD API: generate, list, delete — protected by the same org/resource access middleware as access tokens      
  - Refactored `verifyAccessTokenAccess` and `verifyHeaderTokenAccess` into a shared `verifyTokenResourceAccess`       
  factory, removing ~140 lines of duplication                                                                          
  - Expired token cleanup in `clearStaleData` with proper NULL-safe handling for never-expiring tokens                 
  - PG + SQLite schema, migrations (1.17.0), OpenAPI registration, i18n keys, and frontend UI with token generation,   
  copy-to-clipboard, and delete  

Screenshot From 2026-03-19 15-28-14
Screenshot From 2026-03-19 15-28-22
Screenshot From 2026-03-19 15-28-43

## How to test?
  - [ ] Deploy new version
  - [ ] Enable header token on a resource, set header name, generate token                                             
  - [ ] Verify token grants access via `curl -H "X-Pangolin-Token: <token>"`    
  - [ ] Verify invalid/expired tokens are rejected                                                                     
  - [ ] Verify disabling header token auth blocks token access                                                         
  - [ ] Verify token deletion removes access                                                                           
  - [ ] Verify audit logs show reason 108 for header token access                                                      
  - [ ] Test with both SQLite and PostgreSQL                                                                           
  - [ ] Run fresh migration on clean DB (1.17.0 creates table + column) 

## Notes

This is tested only on my setup with sqlite. I kinda winged the postgresql migrations..


LaurenceJJones:
I’ll leave the main commenting to the founders. My only question is that we technically already solve this with the Links feature. You can create a share link that gives you a URL, or if you go into the details, it shows how to pass an authentication header that Badger can verify. (badger also strips this header before sending downstream because in this implementation badger cannot strip them)

Maybe it just isnt obvious or documented well enough that you can already do this?

crisidev:
The link feature is not exactly what I need I think. My understanding is that it gives you a new url where the parameters are passed as query string and what I need is actually a header I can control. Maybe I am missing something.

LaurenceJJones:
so this header wont work for you?

image


The main issue is this is a feature your saying _you_ need rather than something the community is highly requesting.

crisidev:
Not really because I wouldn't want a different URL. Anyway, if this is not the right direction, happy to close this.

LaurenceJJones:
Just adding this PR here, https://github.com/fosrl/docs-v2/pull/84 cause I dont understand your point about the URL. Maybe the lack of documentation around links created some confusion let me know if this still doesnt satisfy your needs or if im just not fundamentally understanding the url point.

crisidev:
It's probably a very niche problem I have that is not share by a lot of people and I have alternatives (vpn) that do not require this MR.

I have a split DNS and when I am home, reaching my services do not need pangolin. When I am out, some of these services are behind the pangolin SSO and I would like to skip it. Having this header is an easy way to keep my DNS config consistent. 

I hope this explain better my usecase. If this doesn't fit pangolin roadmap, I don't mind closing this and keep using the VPN occasionally.

oschwartz10612:
Would need to think about this a little more, but does the existing header auth work for you at all? 

image

I think this uses the `Authorization` header

crisidev:
Oh right.. I assumed from the documentation it would only work passing user:password@host.. let me try, I'll report back.

crisidev:
Hum.. I cannot really make it work.. you said I could use this with the Authorization header, but I really cannot figure out what I should put the in value of this header :(

LaserPew:
> Hum.. I cannot really make it work.. you said I could use this with the Authorization header, but I really cannot figure out what I should put the in value of this header :(

This is just HTTP Basic Authentication. To build the `Authorization`-header, concatenate the username and password with a colon (`username:password`), Base64‑encode the result, then prefix it with `Basic `.

You can also generate the value with any Basic Auth generator (many are available online), e.g. https://it-tools.tech/basic-auth-generator.

For example, if you were to use `test` as the user and `1234` as the password and enable Header Authentication on your Pangolin resource, you could make a HTTP request with the header being set to this:
`Authorization: Basic dGVzdDoxMjM0`

AstralDestiny:
Just take user:pass and base64 it as that's what that header does anyways.

crisidev:
Oh, I feel pretty dumb now. Yes, it works. I am closing the MR, apologies for the noise and thanks all for the help.

PR #2711, opened 2026-03-25
fix: show informative message when saving with no targets

### LINKED ISSUE #586: Feature request - Empty target validation
Hi, 

I would like to ask to implement validation of newly added resource. I was a bit tired yesterday and I tried to add new resource. I added IP, port etc and I clicked gazzilion times sve target. Even though target is empty the notification appears **Targets Updated**.

Image

I think it would make sense to validate if anything is entered as target and display notification - please enter target first or sth like that.

Cheers, 
Wojciech

jaydeep-pipaliya:
## Summary
- When saving targets with an empty target list and no removals, the app previously showed a misleading "Settings updated" success notification
- Now shows "No targets to save — Please add at least one target before saving" instead
- Added translation keys `targetNoTargets` and `targetNoTargetsDescription` to `en-US.json`

Fixes #586

## Test plan
- [ ] Navigate to a resource's proxy settings
- [ ] Without adding any targets, click "Save Targets"
- [ ] Verify the toast says "No targets to save" instead of "Settings updated"
- [ ] Add a target, save, and verify "Settings updated" still shows correctly
- [ ] Remove all targets and save — verify deletions still process correctly

### DIFF
--- a/messages/en-US.json
+++ b/messages/en-US.json
@@ -568,6 +568,8 @@
     "targetErrorInvalidPortDescription": "Please enter a valid port number",
     "targetErrorNoSite": "No site selected",
     "targetErrorNoSiteDescription": "Please select a site for the target",
+    "targetNoTargets": "No targets to save",
+    "targetNoTargetsDescription": "Please add at least one target before saving",
     "targetCreated": "Target created",
     "targetCreatedDescription": "Target has been created successfully",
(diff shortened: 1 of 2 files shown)


oschwartz10612:
Thanks for the attempt! We want to make sure we can clear the targets though so I think clicking the button when there are no targets is necessary for now.

PR #2766, opened 2026-04-02
enhance: Systemd newt unit instructions

LaurenceJJones:

## Description

Add a systemd unit option directly from dashboard to prevent copy and paste mistakes

## How to test?

When creating a new newt site, you have to option to select systemd service when unix & macos is selected

image

image


src/components/newt-install-commands.tsx
```diff
@@ -64,6 +64,54 @@ export function NewtSiteInstallCommands({
...
+            ],
+            "Systemd Service": [
+                {
+                    title: t("install"),
+                    command: `curl -fsSL https://static.pangolin.net/get-newt.sh | bash`
+                },
+                {
+                    title: t("envFile"),
+                    command: `# Create the directory and environment file
+sudo install -d -m 0755 /etc/newt
```
oschwartz10612, line 76: why sudo install here? maybe this should be mkdir?

oschwartz10612 (changes requested):

src/components/newt-install-commands.tsx
LaurenceJJones: we use `install` as we can set permissions instead of double

`mkdir .... && chown ...`

PR #2799, opened 2026-04-06
fix: use targetId as row identifier

### LINKED ISSUE #2797: Incorrect target deletion in Proxy tab
### Describe the Bug

When managing targets under the Proxy tab of a resource, deleting a specific target does not remove the selected one. Instead, it consistently removes the last target in the list.
> Health check indicators also appear to be bugged and may not reflect the correct status.

A video is attached demonstrating the issue.

https://github.com/user-attachments/assets/a7a1a8c9-a6cd-42cc-82d0-735c68cdd6d6

### Environment

Running on Pangolin Cloud, no local/self-hosted components involved.

### To Reproduce

1. Open a resource
2. Navigate to the Proxy tab
3. Ensure there are multiple targets listed
4. Click Delete on any target (first, middle, etc.)

### Expected Behavior

The selected target should be removed from the list.

LaurenceJJones:
fix: #2797

## Description

When no `getRowId` function is provided it uses the index of the array for the tanstack table. This causes render issues if the data is nested and causes the problem outline in #2797.

## How to test?

What is odd is that I could not reproduce the issue by creating three targets under the same site. To match the user's setup, I had to create three separate sites, since the dashboard uses the site itself as the data point being displayed.

From what I can see, the correct ID does appear to be deleted. The issue seems to be with the UI not updating properly afterward, likely because the data passed into the render is deeply nested and React is not re-rendering that state correctly.

closed #2798 because pointed to wrong branch 🤦🏻

### DIFF
--- a/src/app/[orgId]/settings/resources/proxy/[niceId]/proxy/page.tsx
+++ b/src/app/[orgId]/settings/resources/proxy/[niceId]/proxy/page.tsx
@@ -678,6 +678,7 @@ function ProxyResourceTargetsForm({
         getPaginationRowModel: getPaginationRowModel(),
         getSortedRowModel: getSortedRowModel(),
         getFilteredRowModel: getFilteredRowModel(),
+        getRowId: (row) => String(row.targetId),
         state: {
             pagination: {
                 pageIndex: 0,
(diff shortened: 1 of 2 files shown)


oschwartz10612:
@LaurenceJJones were you able to replicate the issue and confirm this fixes?

LaurenceJJones:
> @LaurenceJJones were you able to replicate the issue and confirm this fixes?

I tested on cloud to replicate then managed to replicate on my own, and this fixed it. I thought you wanted to replicate on Cloud from our call and then test this fix? maybe I got wrong end of the stick from the call?

oschwartz10612 (approved):

PR #2825, opened 2026-04-10
fix(analytics): prevent countries from getting stuck highlighted on world map

AdnanSilajdzic:

## Description
This fixes an issue on the analytics world map where hovered countries could remain highlighted after the cursor moved away.

I was able to reproduce this on the analytics page and traced it back to the hover implementation in `WorldMap.tsx`.

Before:

https://github.com/user-attachments/assets/268c63d3-9784-43e2-bde9-e8cfd4778e06

After applying this fix:

https://github.com/user-attachments/assets/e7a5ba94-2e49-4fe9-b019-dac927079dac

## Cause

The previous implementation moved the hovered SVG country path to the end of the DOM during hover so it would render on top of neighboring countries.

That helped visually, but it also seems to interfere with hover cleanup in some cases, which can leave countries stuck in the highlighted state.

## Fix

Instead of reordering the original country node during hover, this change renders the hovered country in a separate overlay path above the base map layer.

This keeps the hovered country visually on top while avoiding the sticky hover behavior.

## How to test?
This can be tested by navigating to the `/{orgId}/settings/logs/analytics` and interacting with the map on the bottom of the screen.

### DIFF
--- a/src/components/WorldMap.tsx
+++ b/src/components/WorldMap.tsx
@@ -164,7 +164,7 @@ const countryClass = cn(
 
 const highlightedCountryClass = cn(
     sharedCountryClass,
-    "stroke-2",
+    "stroke-[3]",
     "fill-[#f4f4f5]",
     "stroke-[#f36117]",
     "dark:fill-[#3f3f46]"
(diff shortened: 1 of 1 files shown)


oschwartz10612 (approved):

oschwartz10612:
OMG thanks so much for fixing this! We have not been able to figure it out!

PR #2833, opened 2026-04-11
2736 fix healthcheck on disabled resources

rkvf01:

## Description

-  server/routers/newt/buildConfiguration.ts : Added checks to query now filters on resources.enabled as well as targets.enabled 
- server/routers/resource/updateResource.ts : Added syncNewtsForResource()  to push an immediate sync to connected Newt clients when a resource is enabled or disabled.

## How to test?
Added a test 
[repro-2736.ts](https://github.com/user-attachments/files/26648442/repro-2736.ts)


LaurenceJJones:
Currently `newt/sync` [websocket message is ignored](https://github.com/fosrl/newt/blob/3d2b73d417363d4c83a9a3d191c6b58b18e31485/main.go#L1272-L1416)

So this is leading to [even deleted resources continuing](https://github.com/fosrl/newt/issues/308)

oschwartz10612:
Yeah I dont think this is fixing the issue in the way we want to fix it. Thank you for the attempt though. Ideally we want when the target / resource is updated in the CRUD endpoints to send messages down to the newt to remove / add back the health checks with the existing add / remove endpoints.

rkvf01:
okay. Let me take another stab at it. Now that there is some additional context.

PR #2841, opened 2026-04-12
fix(worldmap): correct topojson feature typing

AdnanSilajdzic:

## Summary
This is a small follow-up to my previously merged world map hover fix: [#2825](https://github.com/fosrl/pangolin/pull/2825).

That earlier change fixed the stuck hover state on the analytics map, but it also introduced a TypeScript error. This PR corrects that typing so the fix passes TypeScript cleanly as well.

There is no intended behavior change here. This is just a typing fix for the already-merged hover-state change that I should have included in the original PR. Sorry about that.

## How to test?
`npm run dev:check`

### DIFF
--- a/src/components/WorldMap.tsx
+++ b/src/components/WorldMap.tsx
@@ -237,7 +237,7 @@ function drawInteractiveCountries(
     return svg;
 }
 
-type WorldJsonCountryData = { properties: { name: string; a3: string } };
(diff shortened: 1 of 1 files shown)


oschwartz10612 (approved):

oschwartz10612:
Thank you should have caught that!

Issue #2239, opened 2026-01-12
Private/CGNAT IPs are incorrectly blocked by GeoIP “Block All” rules


Blacks-Army:
### Describe the Bug

When using GeoIP-based rules in Pangolin, internal/private IP ranges (LAN and CGNAT) can be unintentionally blocked.

For example, if you create a rule to Bypass Auth for your country with high priority, and then a Block All rule for all countries with lower priority, clients using private IP ranges (192.168.x.x / 10.x.x.x / 172.16.x.x) or CGNAT ranges (100.64.0.0/10) or private ipv6 IP ranges cannot reach the service. This happens because these IPs do not have GeoIP mapping, so the “Block All” rule applies.

### Environment

- OS Type & Version: Docker
- Pangolin Version: latest 
- Gerbil Version: latest
- Traefik Version: latest
- Newt Version: latest
- Olm Version: (if applicable)

### To Reproduce

1.	Create a rule with Priority 1: Bypass Auth for a specific country (e.g., USA).
2.	Create a rule with Priority 2: Block All for all countries.
3.	Try accessing the service from a client with a private LAN IP (192.168.x.x / 10.x.x.x / 172.16.x.x) or a CGNAT IP.
4.	Access is blocked.

### Expected Behavior

Private IPs and CGNAT ranges should not be blocked by GeoIP-based rules. Internal clients should be allowed by default or configurable in ACLs without requiring manual exceptions.

Suggestion: 
Automatically whitelist RFC1918/RFC6598/RFC4193 ranges in GeoIP evaluations, or provide an explicit setting to treat private/internal IPs separately.

PR #2843, opened 2026-04-12
Exclude local/private/CGNAT IPs from geo-block rules (fixes  issue #2239)

Blacks-Army:

## Description
Fixes issue #2239

Disclaimer:
This is my first contribution to Pangolin, so please bear with me 😅

This PR updates geo wildcard rule behavior for non-public source IPs.

- `COUNTRY=ALL` with action `BLOCK`: local/private/CGNAT IPs are skipped
- `ASN=ALL` / `ASN=AS0` with action `BLOCK`: local/private/CGNAT IPs are skipped
- Explicit `CIDR` / `IP` rules are unchanged

Why this makes sense:
- local/private/CGNAT IPs are not meaningfully geolocatable
- wildcard geo block rules should target public Internet IPs
- intentional blocking is still possible via CIDR, including:
  - `0.0.0.0/0` (IPv4)
  - `::/0` (IPv6)

## How to test?
1. Add rule `COUNTRY=ALL` or `ASN=ALL`,  action `BLOCK`, send request from `192.168.1.10` --> not blocked by this rule.
2. Add rule `CIDR=192.168.0.0/16`, action `BLOCK` --> IP is blocked.
3. Add rule `CIDR=0.0.0.0/0` (or `::/0` for IPv6), action `BLOCK` -> IP is also blocked.


oschwartz10612:
I think this LGTM. There is some concern that people that were for some reason expecting these ranges to be blocked before are not going to find them unblocked. But we can put something in the release notes.

We also would need to update this logic in the remote nodes. Could you open a companion PR there with the same change?

https://github.com/fosrl/pangolin-node

oschwartz10612 (approved):

Blacks-Army:
@oschwartz10612 Just created a PR: https://github.com/fosrl/pangolin-node/pull/61

> There is some concern that people that were for some reason expecting these ranges to be blocked before are not going to find them unblocked. But we can put something in the release notes.

I understand this concern, but in my opinion local IPs are not related to any country or ASN and therefore should not be blocked by such rules.

If you want to block everything, you could instead block the CIDR range 0.0.0.0/0.

oschwartz10612:
Ignore these tests they are from a mistake on the dev branch

Blacks-Army:
@oschwartz10612 Please let me know if anything else is needed, or if I can expect this patch in the next release.

PR #2900, opened 2026-04-24
(fix): support query parameters in PATH rules

### LINKED ISSUE #839: Issue with blocking URLs with "?" in Pangolin
Hi everyone,

First of all, thanks for the great work on Pangolin — I’m new to the project and really enjoying it so far!

Today I tried to add a rule to block specific login URLs for my self-hosted services, starting with Audiobookshelf. I’d like to block all requests to:

/audiobookshelf/login/?autoLaunch=0

This URL allows bypassing the automatic redirect to the OAuth provider, which I neither want nor need. I’d like to prevent direct access via this query parameter.

I encountered the same issue with Immich:

/auth/login?autoLaunch=0

Unfortunately, I couldn’t get it to work — even when trying with wildcards.
Is there a recommended way to block requests with specific query parameters like this in Pangolin?

Thanks in advance for your help!

sidd190:
<h2>fix: support query parameters in PATH rules (fixes #839)</h2>
<p>Closes #839</p>

[Screencast from 2026-04-24 23-48-12.webm](https://github.com/user-attachments/assets/a7ac7609-cd46-467b-82bd-84eb4df2ef32)

<hr>
<h3>Problem</h3>
<p>PATH rules had no awareness of query strings. A pattern like <code>/auth/login?autoLaunch=0</code> was treated as invalid by the validator, and even if stored, the matching logic in <code>verifySession.ts</code> only compared the path segment - the <code>?</code> and everything after it was silently ignored.</p>
<p>This made it impossible to block URLs that bypass OAuth redirects via query parameters, such as:</p>
<ul>
<li><code>/audiobookshelf/login/?autoLaunch=0</code></li>
<li><code>/auth/login?autoLaunch=0</code></li>
</ul>
<hr>
<h3>Changes</h3>
<p><strong><code>server/routers/badger/pathMatching.ts</code></strong> -> new file containing <code>isPathAllowed</code> extracted into its own module. Extends the function with a query parameter and adds query string matching logic: parses <code>key=value</code> pairs from the pattern, checks each required param exists in the request query, and supports <code>*</code> glob wildcards in values (e.g. <code>/path?key=*</code>). Multiple params via <code>&amp;</code> are supported. The <code>query</code> field was already being sent by badger and parsed by the schema - it just wasn't being used for rule matching. Another reason was modifying it directly in the `verifySession.ts` file was creating a circular dependency between it and the tests.</p>
<p><strong><code>server/routers/badger/verifySession.ts</code></strong> -> passes the parsed query object through <code>checkRules</code> into <code>isPathAllowed</code>. Re-exports <code>isPathAllowed</code> from <code>pathMatching.ts</code> to preserve the existing public API.</p>
<p><strong><code>server/lib/validators.ts</code></strong> -> updated <code>isValidUrlGlobPattern</code> to split on <code>?</code> before validating, then validate the path and query parts separately with appropriate character sets. The frontend already imports this validator, so the rule creation UI now accepts query-containing patterns without any UI changes.</p>
<p><strong>Tests : </strong> New tests added, tests for the validator and tests <code>isPathAllowed</code> ( for path matching + the new query param matching) and <code>isIpInRegion</code> in files <code>validators.test.ts</code> and <code>verifySession.test.ts</code> </p>
<hr>
<h3>Behaviour</h3>

<p>Patterns without a <code>?</code> continue to work exactly as before.</p></body></html>


AstralDestiny:
@sidd190 But did you verify that it functions against traefik which is handing this?

sidd190:
> @sidd190 But did you verify that it functions against traefik which is handing this?

Hellooo @AstralDestiny , I didn't. I think even the implementation would need a few updates from me just yet. This is a very raw PR for now. Should I convert it to draft? I would like for it to not be reviewed for a while yet.

AstralDestiny:
Yeah a draft is fine.

PR #2932, opened 2026-04-29
Feature/local resource healthchecks

pizzaandcheese:

## Description
This change adds healthcheck support for local sites in Pangolin.

## How to test?

 You need a Pangolin instance with at least one Local site configured — a site where the target is directly reachable by the Pangolin server over the network (not tunnelled through Newt).

Test 1: Healthcheck turns healthy

1. Go to a resource attached to a local site, open the proxy settings, and click the healthcheck button on a target
2. Configure a simple HTTP check pointing at a running service (e.g. GET / expecting 200)
3. Enable it and save
4. Within a few seconds the target should transition from Unknown → Healthy
5. The resource card should also reflect Healthy

Test 2: Healthcheck detects a failure

1. With the healthcheck running and showing Healthy, stop or block the target service
2. Within one polling interval the target should transition to Unhealthy
3. Restart the service — it should return to Healthy

Test 3: TCP mode

Repeat tests 1 and 2 using TCP mode instead of HTTP to verify the TCP probe path works independently

Test 4: Config options

- Set a custom interval (e.g. 10s) and verify the polling frequency changes
- Set an unhealthy interval (e.g. 60s) and verify it polls less aggressively once unhealthy
- Set a healthy/unhealthy threshold > 1 and verify the target doesn't flip status until that many consecutive results are seen

Test 5: WireGuard sites are unaffected

On a WireGuard site, verify the healthcheck button still does not appear and no probing happens

Test 6: Newt sites are unaffected

On a Newt site, verify healthchecks still work exactly as before — the Newt agent should still be doing the probing, not the server

Test 7: Enable/disable toggle
    
Disable an active healthcheck and verify the status resets to Unknown and polling stops. Re-enable it and verify it resumes from Unknown → Unhealthy → Healthy

### DIFF
--- a/server/routers/target/index.ts
+++ b/server/routers/target/index.ts
@@ -4,3 +4,4 @@ export * from "./deleteTarget";
 export * from "./updateTarget";
 export * from "./listTargets";
 export * from "./handleHealthcheckStatusMessage";
+export * from "./localHealthChecker";
--- a/server/routers/target/localHealthChecker.ts
+++ b/server/routers/target/localHealthChecker.ts
@@ -1,0 +1,349 @@
+import { db, primaryDb, sites, targetHealthCheck, targets } from "@server/db";
+import logger from "@server/logger";
+import { and, eq, inArray } from "drizzle-orm";
+import {
+    fireHealthCheckHealthyAlert,
+    fireHealthCheckUnhealthyAlert
+} from "#dynamic/lib/alerts";
+import * as net from "net";
+
+// Local healthcheck pollers run server-side for sites of type "local",
+// because there is no Newt agent on the remote side to perform the probes.
+// We re-use the targetHealthCheck table and the same alert helpers used by
+// the Newt-driven flow so the rest of the system (UI, alerts, history) is
+// unchanged.
+
+const LOCAL_HC_TICK_MS = 5 * 1000; // re-evaluate which checks are due every 5s
+const DEFAULT_INTERVAL_S = 30;
+const DEFAULT_TIMEOUT_S = 5;
+
+let localHealthCheckerInterval: NodeJS.Timeout | null = null;
+
+// In-memory state for each local healthcheck so we can implement
+// healthy / unhealthy thresholds and per-check intervals without hammering
+// the database with extra columns.
+type RunState = {
+    nextRunAtMs: number;
+    consecutiveHealthy: number;
+    consecutiveUnhealthy: number;
+    lastStatus: "unknown" | "healthy" | "unhealthy";
+    inFlight: boolean;
+};
+
+const runState = new Map<number, RunState>();
+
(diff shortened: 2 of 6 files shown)


AstralDestiny:
So pangolin itself does the healthchecks at that point or you leverage traefik's own built in healthchecks and query that?

pizzaandcheese:
Currently, the Pangolin process runs the health checks. This is in line with how healthchecks are already done with Gerbil.

This can be expanded to cover Wireguard hosts in the same way.

AstralDestiny:
Actually pangolin doesn't do the healthchecks it's newt that does them right now.

oschwartz10612:
Hey thank you so much for this PR! I think right now we are going to keep the hc functionality in newt. I appreciate you opening this though!

keonramses:
Hi @oschwartz10612, thanks for taking a look at this! I completely understand wanting to centralize health check functionality in Newt. However, I’d love for the team to reconsider this approach for local sites. This feature has been highly requested by the community in issues like #1835 and #1873. A major driving factor here is that many users experience performance bottlenecks with Newt (as tracked in #512). For environments where the Pangolin server can reach the target directly, allowing native server-side health checks removes that unnecessary overhead and provides a much more stable experience. Since the PR is already written and limits the scope strictly to local/server-reachable sites without breaking Newt's existing flow, would you be open to revisiting this as an optional/opt-in feature?

oschwartz10612:
1.19 is locked but willing to take a look for a different release

keonramses:
Thank you.

cookie-monster-bit:
+1 on this — healthchecks for Local sites would be a great addition. Right now this capability only exists for Newt-tunnelled sites, so anyone running Local resources (directly reachable by the Pangolin server) has no way to get health status on their targets. A way to overcome this is by running newt along side Pangolin, but this would cause unnecessary overhead. Adding this would close that gap nicely. Thanks you for your guy's hard work on Pangolin!

AstralDestiny:
You can still get it to function via,

```yml
services:
 gerbil:
  networks:
   default:
    aliases: 
     - pangolin.domain.com # We tell newt it can find this container via docker networking over going out to the world and back in.
 pangolin:
  networks:
   - default
# We don't touch Traefik network.

services:
  newt:
    image: fosrl/newt
    restart: unless-stopped
    depends_on:
      pangolin:
        condition: service_healthy
      gerbil:
    environment:
      - PANGOLIN_ENDPOINT=https://pangolin.domain.com
      - NEWT_ID=
      - NEWT_SECRET=
    networks:
      - default
      - SomeOtherNet

networks:
  default:
    name: pangolin
    enable_ipv6: true
  SomeOtherNet: # docker network create <name>
   external: true
```
So the tunnel is effectively sub microsecond/nanosecond for it's connection and all internal to the kernel which can do multi gig even if the host only can do a gig itself.

PR #2980, opened 2026-05-03
Auto-create roles referenced in blueprints

rinseaid:

### Preface

This PR was developed with Claude Code (claude-opus-4-6). I drove the direction and tested everything, but Claude did all of the actual coding. I reviewed, tested, and validated all changes.

## Description

When a blueprint references a role that doesn't exist yet, create it automatically instead of failing or silently dropping the association.

- Proxy resource blueprints (`sso-roles`) previously threw `Role not found`, blocking the entire blueprint apply. Now auto-creates the role.
- Client resource blueprints (`roles`) previously ran a `WHERE IN` query and silently ignored any role names that didn't match. Now auto-creates missing roles before associating them.
- Auto-created roles receive the three default permissions: `getOrg`, `getResource`, `listResources`.

Blueprints are the declarative source of truth for resource configuration. Requiring roles to be pre-created through the UI before they can be referenced in a blueprint breaks the declarative workflow and adds a manual step that's easy to forget. This is especially relevant when integrating with an external IdP where SSO group-to-role mappings should just work on first apply.

## How to test?

- Apply a blueprint with `sso-roles` referencing a role that does not exist. Verify the role is created with default permissions and the resource association succeeds.
- Apply a blueprint with `roles` (client resource) referencing a role that does not exist, for both new and existing resources. Verify the same.
- Apply a blueprint referencing a role that already exists. Verify no duplicate is created and the existing role is associated normally.
- Verify auto-created roles appear in the admin UI with the correct permissions.

### DIFF
--- a/server/lib/blueprints/clientResources.ts
+++ b/server/lib/blueprints/clientResources.ts
@@ -3,6 +3,7 @@ import {
     clientSiteResources,
     domains,
     orgDomains,
+    roleActions,
     roles,
     roleSiteResources,
     Site,
@@ -19,6 +20,7 @@ import { sites } from "@server/db";
 import { eq, and, ne, inArray, or, isNotNull } from "drizzle-orm";
 import { Config } from "./types";
 import logger from "@server/logger";
+import { defaultRoleAllowedActions } from "@server/routers/role/createRole";
 import { getNextAvailableAliasAddress } from "../ip";
 import { createCertificate } from "#dynamic/routers/certificates/createCertificate";
 
@@ -335,8 +337,7 @@ export async function updateClientResources(
             }
 
             if (resourceData.roles.length > 0) {
-                // Re-add specified roles but we need to get the roleIds from the role name in the array
-                const rolesToUpdate = await trx
+                const existingRoles = await trx
                     .select()
                     .from(roles)
                     .where(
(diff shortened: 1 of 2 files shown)


LaurenceJJones:
Maybe im overthinking it a little but if the role doesnt exist before the blueprint applies then nobody is applied to the group so even if we generate the role, nobody would still be granted access?

So a user would still need to login and apply the role to the users, then at that point simply creating the role becomes trivial? unless you forsee that a role will be generated with some randomness but then how would you know which roles you need to apply to each users without just using the integration api?

Sorry brain dumping to trying to understand that even if we unlock the apply to generate a role what usefulness does it provide

AstralDestiny:
My brain was thinking would be nice to make a role in a seperate thing and be able to authorize or tie it to idp's from within blueprints but also maybe setting it within might cause duplicates if you're not careful? Though without a way to auto list them to an idp then not sure wherethis goes.. 

brain dumping like Loz hehe

rinseaid:
@LaurenceJJones - this actually already works through auto provisioning. If you create a Pangolin role, and then login with an OIDC user that has a claim for that role, they are auto-added and can access the resource. 

@AstralDestiny - that is quite a valid point. It appears that the roles table doesn't have any constraint to prevent role duplication. I think that issue is deserving of its own PR, as it really needs a new unique index in the roles table and associated DB migration. This PR doesn't create the problem, but it obviously increases its likelihood.

AstralDestiny:
Would also like to have auto provisioning to be able to be define auto-provisioned settings from blueprints too..

rinseaid:
I put up a separate PR for the unique index on `(orgId, name)` in the roles table: #3013. It needs a data migration to deduplicate any existing rows before the constraint can be created, so it made more sense as its own change.

rinseaid:
Just updating that I decided to close #3013 (the unique index migration) as it was more complex than it needed to be. The duplicate risk AstralDestiny flagged still applies here, so an approach still needs to be considered if this moves forward.

oschwartz10612 (approved):

oschwartz10612:
I think this makes sense thank you! :}

rinseaid:
Awesome, thank you!

PR #3013, opened 2026-05-06
Add unique index on roles (orgId, name)

rinseaid:

### Preface

This PR was developed with Claude Code (claude-opus-4-6). The migration logic, edge case handling, and scoped dedup approach went through multiple review rounds (11 agents across 2 passes) that caught several bugs that would have caused data loss in production, including a type coercion bug where postgres returns `MIN()` as a string instead of an integer, and a connection pool race where raw `BEGIN`/`COMMIT` doesn't pin a connection. Both PostgreSQL and SQLite migrations were tested against real databases with intentionally messy data (triple duplicates, PK conflicts, legitimate duplicate rows in FK tables that need to survive). I reviewed, tested, and validated all changes.

## Description

The `roles` table has no unique constraint on `(orgId, name)`. Multiple code paths use a select-then-insert pattern that can produce duplicate rows under concurrent requests: the `createRole` API handler and the blueprint auto-create logic in #2980.

This adds a versioned migration (1.19.0, both PG and SQLite) that cleans up any existing duplicates before creating the index. For each duplicate group (same orgId + name), the migration keeps the lowest roleId and repoints all foreign key references from the duplicates to the survivor across the 7 tables that reference `roles.roleId` (userOrgRoles, roleSiteResources, roleActions, roleSites, roleResources, userInviteRoles, roleClients). Any rows that become identical after the repoint are deduplicated. The duplicate roles are then deleted and the unique index is created.

Schema files are updated to declare the `uniqueIndex` so Drizzle stays in sync with what the migration creates.

## How to test?

- Deploy against a database with duplicate roles (same orgId + name). Verify duplicates are merged, the lowest roleId survives, and all FK references are repointed.
- Verify no orphan references remain in any of the 7 FK tables.
- Verify legitimate duplicate rows in FK tables for unrelated roles are not collapsed.
- Attempt to insert a duplicate role (same orgId + name). Verify it is rejected.
- Insert a role with the same name in a different org. Verify it succeeds.

### DIFF
--- a/server/db/pg/schema/schema.ts
+++ b/server/db/pg/schema/schema.ts
@@ -11,6 +11,7 @@ import {
     serial,
     text,
     unique,
+    uniqueIndex,
     varchar
 } from "drizzle-orm/pg-core";
 
@@ -443,7 +444,9 @@ export const roles = pgTable("roles", {
     sshSudoCommands: text("sshSudoCommands").default("[]"),
     sshCreateHomeDir: boolean("sshCreateHomeDir").default(true),
     sshUnixGroups: text("sshUnixGroups").default("[]")
-});
+},
+    (t) => [uniqueIndex("idx_roles_orgId_name").on(t.orgId, t.name)]
+);
 
 export const userOrgRoles = pgTable(
     "userOrgRoles",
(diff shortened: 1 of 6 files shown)


rinseaid:
Closing - conflicts with 1.18.4 on main and a migration this complex is probably beyond what I should be submitting as an outside contributor. A simpler fix at the insert sites might be a better fit if this becomes a real issue.

PR #3058, opened 2026-05-13
1.18.4

oschwartz10612:
- **allow editing self and owner user roles**
- **Pick the most specific domain**
- **Confirm delete of share links**
- **Use the right param for user**
- **Add global hide_powered_by and make it backward**
- **improve targets round robin warning**
- **Show when a domain is config managed**


miloschwartz (approved):

AstralDestiny:
> improve targets round robin warning

So when do we get stuff like Power of 2 and such?

oschwartz10612:
Not quite yet! :{

This is just a warning on the cloud

Issue #3041, opened 2026-05-09
Client connections and LAN resource unavailability after updates


GisaLab:
### Describe the Bug

Hello
After upgrading the Pangolin server to v1.18 and Newt to v1.12.5, several client machines remain disconnected (including one machine upgraded to pangolin-cli 0.8.0 and then recreated as 0.8.1), and the LAN site sometimes becomes unavailable (disconnections observed lasting 7 and 23 minutes). Since the Newt update to v1.12.5, the site indicates it is connected, but no site resources are accessible.

### Environment

- OS Type & Version: Ubuntu 24.04.4 LTS
- Pangolin Version: v1.18.3
- Gerbil Version: 1.4.0
- Traefik Version: v3.6.16
- Newt Version: v1.12.5
- Olm Version: 1.5.0

### To Reproduce

1. **Prerequisites**
   - Pangolin server in `v1.18.3`.
   - Client(s) using `pangolin-cli` (e.g. `0.8.0` then `0.8.1`).
   - Newt installed on gateway LAN (updated to `v1.12.5`).
   - Olm in `1.5.0` (or `1.4.4` for regression testing).
2. **Steps**
   1. Update the Pangolin server to `v1.18.3`.
   2. Update (or install) `pangolin-cli` on a client machine (e.g. `0.8.0`), observe the status in the interface.
   3. If the machine remains disconnected, delete and then recreate the machine in `pangolin-cli 0.8.1`.
   4. Update Newt on the LAN gateway to `v1.12.5`.
   5. Check access to LAN site resources (pages, services exposed via Traefik / Newt).
3. **Observed Behavior**
   - Client machine remains in **disconnected status** despite Olm CLI `v1.5.0` or `v1.4.4`.
   - After updating Newt → `v1.12.5`, the site displays **connected status** but **no resources are accessible**.
   - Newt logs show repeated ICMP ping failures and loss of connection to the server (extracts below).
4. **Relevant log excerpts**
```text
Newt:
May 09 16:26:10 newt-gateway-lan systemd[1]: Started newt.service - Newt.
May 09 16:26:10 newt-gateway-lan newt[234]: INFO: 2026/05/09 16:26:10 Newt version 1.12.5
May 09 16:26:11 newt-gateway-lan newt[234]: INFO: 2026/05/09 16:26:11 Loading config from: /root/.config/newt-client/config.json
[...]

Issue #2951, opened 2026-05-01
--holepunch=false (relay mode) fails when UDP 21820 is blocked — server still requires recent hole punch for registration


z10n-dev:
### Describe the Bug

When UDP port 21820 is blocked by a firewall (e.g. school or corporate network), the `--holepunch=false` flag does not work as expected. Even with holepunch disabled on the client, the server still requires a successful hole punch within the last 5 seconds before it will respond to `olm/wg/register`. This makes it impossible to use relay mode on networks that block UDP 21820.

**Actual behavior**
Server-side validation in `handleOlmRegisterMessage` checks `if (now - (client.lastHolePunch || 0) > 5 && sitesCount > 0)` and skips the register regardless of whether the client requested relay mode.

**Additional context**
- The connection works fine on a hotspot (UDP 21820 unblocked)
- Port 21820 UDP is hardcoded in Gerbil and not yet configurable via config.yml (`clients_start_port` does not affect Gerbil's actual listening port)
- This is a regression from pre-0.8.0 behavior, where relay fallback worked silently when hole punch failed
- tcpdump on the server confirms hole punch packets from the school network never reach port 21820, but do reach 51820

### Environment

- Pangolin server: 1.18.1
- Olm client: 0.8.0
- OS: Arch Linux (kernel 7.0.2-arch1-1)
- Network: School WiFi — UDP 21820 blocked, UDP 51820 allowed (confirmed via tcpdump)

### To Reproduce

1. Block outbound UDP port 21820 (or be on a network that does, e.g. school/corporate WiFi)
2. Run: `pangolin up --holepunch=false`
3. The websocket connects successfully, but the server keeps responding with:
   `Client last hole punch is too old and we have sites to send; skipping this register`
4. The client times out and disconnects — relay mode never establishes

### Expected Behavior

When `--holepunch=false` is passed, the server should accept the registration without requiring a recent hole punch and fall back to relay mode, just as it did in versions prior to 0.8.0.

AstralDestiny:
Relay is done via 21820 Out to the pangolin stack then it connects down to site over you going directly.

z10n-dev:
So there is no other way, than using my Hotspot, when the port 21820 is blocked in school wifi?

oschwartz10612:
Not right now sorry. Relay over UDP is the only option though I know some work is being done on websocket relaying which we might be able to take.

jecabe:
I don't think this is only related to the server configuration.

[...]

Issue #2901, opened 2026-04-24
Adding a private resource breaks hole-punch for already-connected clients; cache never repopulates


andrewixl:
### Describe the Bug

Creating any private resource (siteResources) and once created, if an Olm client is registered (connected) puts the client into a permanent "last hole punch is too old" loop. The client never reaches connected state, and no amount of reconnecting, restarting, or deleting/recreating the resource restores it.

### Environment

- Ubuntu 24.04
- Pangolin: ee-1.17.1
- Gerbil: 1.31
- Traefik: 3.6
- Newt: 1.11.0
- Olm Windows: 0.9.0

### To Reproduce

Reproduction:
1. No private resources.
2. Create a new client. Pair it with a Windows Pangolin client. Confirm connected.
3. Add ONE private resource pointing at any LAN IP.
4. Immediately: logs show "handleOlmServerPeerAddMessage: No endpoint found for client N"
5. Client enters permanent "Client last hole punch is too old" loop
6. clientSitesAssociationsCache rows for that client have endpoint=NULL, publicKey=NULL
7. clients.lastHolePunch remains NULL
8. Deleting the resource restored the client to connectivity.

### Expected Behavior

Expected behavior should allow the client to register and connect to the private resources.

Byolock:
From my personal experience and some other comments here, that could be a DNS Issue. It looks like you are using windows as a client? Try adding

`192.168.1.10 pangolin.yourdomain.com` 

to C:\Windows\System32\drivers\etc\hosts for a quick and easy test. Replace the IP and domain to match your setup. 

If it works after this you have a DNS issue. I've read reports that blocking outbound DNS for clients in your firewall does this. I just experienced it also happens when you rely on your local DNS Server in a Split DNS Setup. Every device in my lan uses an adguard dns server and resolves my "pangolin.mydomain.com" record correctly to my lan address of the pangolin installation. But using public DNS "pangolin.mydomain.com" resolves to a different address. I needed to put a A-record for "pangolin.mydomain.com" to my local address in the public DNS and then it worked, even though I was only testing everything inside my lan.

Issue #2343, opened 2026-01-26
iOS and macOS app stuck at registering


mrprofessork:
### Describe the Bug

iOS and macOS clients stuck at registering when trying to connect. The pangolin logs show: 
`client last hole punch is too old and we have sites to send; skipping this register`

After removing the private resources and archiving all user client devices, it connects. as soon as I re add a private resource, CIDR or IP address, it stops connecting and goes back to the same error. 

### Environment

- OS Type & Version: (e.g., Ubuntu 22.04) ubuntu 24.04.2
- Pangolin Version: 1.15.1
- Gerbil Version:1.3.0
- Traefik Version: v3.4.1
- Newt Version: v1.9.0
- Olm Version: (if applicable)

### To Reproduce

Create private resource, try to connect ios or macos client, if it gets stuck at registering, remove private resource and archive clients and it will connect.

### Expected Behavior

In previous versions, I was able to connect the macos client and get access to private resources remotely.

mrprofessork:
Restarting Newt seems to have resolved this.

xdonu2x:
I have the same problem. Removing private resource and archive client and rstarting Newt didn't resolve this. 
Any other ideas are appreciated.

mrprofessork:
You can try this:
Open UDP port 21820 open on your VPS and in the docker compose. If you have upgraded from older versions of the stack this was not the default. Also please make sure you are on the latest version of gerbil 1.3.0. To do this in your docker compose file just add this line under the ports section: "- 21820:21820/udp"

xdonu2x:
Thank you mrprofessork, need to add the 21820 port to my docker compose file. My iphone can now connect but I still can't access the CIDR private resource. I have explicitly added my user (also the admin) to the private resource.

mrprofessork:
Ok No problem.

Do you have the CIDR resource like this? 192.168.1.0/24 

Also try removing the explicit user from the access policy and if you have more than 1 site, make sure the correct site is selected.

xdonu2x:
Yes, CIDR 192.168.1.0/24. I only have one site. I even tried host private resource, still can't seem to access it

mrprofessork:
Hmm, not sure, maybe head over to their discord and see if someone there may be able to help. Without actually seeing the setup I won't be able to offer more help. Sorry about that.

xdonu2x:
Thanks again.

PR #3068, opened 2026-05-14
Return error response for stale hole-punch rejection

rinseaid:

### Preface

This PR was developed with Claude Code (claude-opus-4-6). I identified the bug through live debugging of a broken Olm tunnel after a Pangolin server restart, traced the root cause through the code, and directed the fix. Claude wrote the patch. I reviewed and validated all changes.

## Description

When `handleOlmRegisterMessage` rejects a registration because `lastHolePunch` is stale (more than 5 seconds old and the client has sites to receive), it performs a bare `return` with no response to the client. The client has no way to know its registration was rejected or why, so it retries blindly until it exhausts its retry budget (20 attempts at 2-second intervals), then gives up silently.

This is particularly problematic during server restarts: the server's in-memory `lastHolePunch` timestamps reset, so the first registration attempt after reconnect is always rejected. Combined with the Olm re-registration bug (fosrl/olm#122), this creates a double failure where the client either never re-registers or re-registers into a silent rejection loop.

**Fix:** Send an `olm/error` response with a new `STALE_HOLE_PUNCH` error code before returning. This uses the existing `sendOlmError` pattern that other rejection paths in the same function already use. The Olm client's `handleOlmError` handler will log the error and surface it via the API server status endpoint, giving users visibility into the failure.

Related issues: #2951, #2901, #2318, #2343, #3041, #1526

Companion PR: fosrl/olm#122

## How to test?

1. Connect an Olm client with at least one site configured.
2. Restart the Pangolin server.
3. Watch Olm client logs after reconnect.
4. **Before fix:** Client retries registration 20 times with no feedback, then silently gives up. Server logs warn about stale hole punch but client sees nothing.
5. **After fix:** Client receives `STALE_HOLE_PUNCH` error and logs it. The error is visible in the Olm status API.


LaurenceJJones:
Hey 👋🏻 

From our [contributing guidelines](https://docs.pangolin.net/development/contributing) you should base your branch on upstream dev and when opening a PR aim it at that branch instead of main.

Furthermore your PR inroduces a new github workflow which is creating a docker image based on your naming, if this was for testing then please cherry pick actual minimal changes that are needed for your change rather than adding mirgrations (which may be because you did start on dev but aiming at main).

rinseaid:
Hi @LaurenceJJones, thanks for letting me know and apologies for the noise! I had other changes on the branch from a separate build and didn't clean them out before opening the PR. I'll close this and reopen with a clean branch off dev containing only the two relevant files! New PR: https://github.com/fosrl/pangolin/pull/3069

Issue #2318, opened 2026-01-24
Connection Failure: "Client last hole punch is too old" on Android and CLI due to SQLITE_CONSTRAINT_FOREIGNKEY


r3nor:
### Describe the Bug

I am unable to establish a connection from Android or CLI after the first successful attempt. The server appears unable to clear the previous session, leaving the client in a broken state where it constantly fails with "Client last hole punch is too old."

I observed a FOREIGN KEY constraint failed error in the logs immediately preceding the connection issues. It appears the server fails to clean up expired sessions, preventing new connections from registering correctly.

**Root Cause Log (Pangolin Container):**

The server crashes when trying to clear expired sessions:

```
pangolin  | 2026-01-24T18:11:28+00:00 [warn]: Error clearing expired sessions: FOREIGN KEY constraint failed
pangolin  | Stack: SqliteError: FOREIGN KEY constraint failed
pangolin  |     at PreparedQuery.run (/app/node_modules/src/better-sqlite3/session.ts:132:20)
pangolin  |     at QueryPromise.resultCb (/app/node_modules/src/sqlite-core/session.ts:185:61)
pangolin  |     at QueryPromise.execute (/app/node_modules/src/sqlite-core/session.ts:31:15)
pangolin  |     at QueryPromise.then (/app/node_modules/src/query-promise.ts:31:15) {"code":"SQLITE_CONSTRAINT_FOREIGNKEY"}
```

**Symptom Log (Connection Attempt):**

Following the SQL error, the client attempts to connect but gets rejected due to key mismatches and stale hole punch data:

```
pangolin  | 2026-01-24T18:19:00+00:00 [info]: WebSocket connection established - OLM ID: 3xxrgre3mlg57ia
pangolin  | 2026-01-24T18:19:01+00:00 [info]: Handling register olm message!
pangolin  | 2026-01-24T18:19:01+00:00 [info]: Public key mismatch. Updating public key and clearing session info...
pangolin  | 2026-01-24T18:19:01+00:00 [warn]: Client last hole punch is too old and we have sites to send; skipping this register
...
```

**Observations**:

- The log shows a Public key mismatch followed by repeatedClient last hole punch is too old warnings.
- This occurs on both platforms (Android and Linux).

### Environment

- OS Type & Version: (e.g., Ubuntu 22.04): Linux Debian 13 Trixie
- Pangolin Version: latest tag
- Gerbil Version: latest tag
- Traefik Version: v3.1 tag
[...]

Issue #1526, opened 2025-09-23
Hole punch process failing - prevents tunnel establishment and all connectivity


devarmada:
## Problem
Newt client connects via WebSocket but tunnel never establishes because hole punch process fails, leaving site.endpoint null and blocking WireGuard configuration delivery.

## Environment
• **Pangolin Server:** 1.10.2 in Podman on Debian VPS
• **Gerbil:** 1.2.1 in Podman  on Debian VPS  
• **Newt Client:** 1.5.0 on Raspberry Pi 4, Raspbian Bullseye (ARM64)
• **Network:** Newt on local network 172.19.1.0/24, Pangolin on public VPS

## Root Cause
Hole punch process fails → site.endpoint remains null → WireGuard config blocked → no tunnel

Flow:
1.  Newt connects via WebSocket
2.  Newt requests newt/wg/get-config
3.  Pangolin checks site.endpoint - it's null
4.  Returns early, no WireGuard config sent
5.  No tunnel established

## Evidence

# Pangolin logs - can't find endpoint
`2025-09-23T10:05:32.524Z [warn]: Site 2 has no endpoint, skipping`

# Newt logs - can't get WireGuard config  
`Sep 23 12:05:50 dente newt[408]: INFO: SendMessageInterval timed out after 10 attempts for message type: newt/wg/get-config`

## Impact
• HTTP resources: 504 Gateway Timeout
• TCP resources: Non-functional  
• All tunnel-dependent features broken
• Core Pangolin functionality unusable

## Versions Tested
• Pangolin 1.10.2 + gerbil 1.2.1 + newt 1.5.0

The hole punch endpoint `/api/v1/gerbil/update-hole-punch` is never successfully called, preventing the site.endpoint field from being set, which blocks all tunnel functionality.

AstralDestiny:
You have wireguard port open? and does your newt succeed even getting to icmp ping tests?

oschwartz10612:
^^ port 51820 and port 21820 https://docs.digpangolin.com/manage/clients/add-client#prerequisites

devarmada:
I have missed that resource, sorry.
Port 51820 was already there, I've added port 21820
The VPS has no firewall, BTW

These are the logs from the newt client
```
-- Journal begins at Mon 2025-09-22 13:12:38 CEST. --
Sep 23 15:42:23 dente systemd[1]: newt.service: Succeeded.
[...]

PR #3069, opened 2026-05-14
Return error response for stale hole-punch rejection

rinseaid:

### Preface

This PR was developed with Claude Code (claude-opus-4-6). I identified the bug through live debugging of a broken Olm tunnel after a Pangolin server restart, traced the root cause through the code, and directed the fix. Claude wrote the patch. I reviewed and validated all changes.

## Description

When `handleOlmRegisterMessage` rejects a registration because `lastHolePunch` is stale (more than 5 seconds old and the client has sites to receive), it performs a bare `return` with no response to the client. The client has no way to know its registration was rejected or why, so it retries blindly until it exhausts its retry budget (20 attempts at 2-second intervals), then gives up silently.

This is particularly problematic during server restarts: the server's in-memory `lastHolePunch` timestamps reset, so the first registration attempt after reconnect is always rejected. Combined with the Olm re-registration bug (fosrl/olm#123), this creates a double failure where the client either never re-registers or re-registers into a silent rejection loop.

**Fix:** Send an `olm/error` response with a new `STALE_HOLE_PUNCH` error code before returning. This uses the existing `sendOlmError` pattern that other rejection paths in the same function already use. The Olm client's `handleOlmError` handler will log the error and surface it via the API server status endpoint, giving users visibility into the failure.

Related issues: #2951, #2901, #2318, #2343, #3041, #1526

Companion PR: fosrl/olm#123

## How to test?

1. Connect an Olm client with at least one site configured.
2. Restart the Pangolin server.
3. Watch Olm client logs after reconnect.
4. **Before fix:** Client retries registration 20 times with no feedback, then silently gives up. Server logs warn about stale hole punch but client sees nothing.
5. **After fix:** Client receives `STALE_HOLE_PUNCH` error and logs it. The error is visible in the Olm status API.

### DIFF
--- a/server/routers/olm/error.ts
+++ b/server/routers/olm/error.ts
@@ -86,6 +86,11 @@ export const OlmErrorCodes = {
     TERMINATED_BLOCKED: {
         code: "TERMINATED_BLOCKED",
         message: "This session was terminated because access was blocked."
+    },
+    STALE_HOLE_PUNCH: {
+        code: "STALE_HOLE_PUNCH",
+        message:
+            "Hole punch is stale. Ensure the server is reachable on the configured UDP port."
     }
 } as const;
 
--- a/server/routers/olm/handleOlmRegisterMessage.ts
+++ b/server/routers/olm/handleOlmRegisterMessage.ts
@@ -306,12 +306,12 @@ export const handleOlmRegisterMessage: MessageHandler = async (context) => {
 
     // this prevents us from accepting a register from an olm that has not hole punched yet.
     // the olm will pump the register so we can keep checking
-    // TODO: I still think there is a better way to do this rather than locking it out here but ???
     if (now - (client.lastHolePunch || 0) > 5 && sitesCount > 0) {
         logger.warn(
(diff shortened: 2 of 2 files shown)


oschwartz10612:
Hi! Thanks so much for your attempt on this change as well. I dont think this will work either as it is.

The reason is that the register message is intentionally pumped by olm so the handler runs multiple times. This gives it a chance to still register if a ws message is lost, or if the holepunch is not logged into the db yet. So if we were to return the error on the first request it would not give the pumping the opportunity to do its work. 

For proper feedback to the user what we should do on the olm side is include a flag in the message if its the last message in the send message interval function. This could be picked up by the pangolin server and we could return this error to the client if it still has not holepunched. Then, we would also need the implementation on the olm side to make sure that we encode this error code correctly and in each client (windows, cli, apple, android) we surface the error correctly.

rinseaid:
@oschwartz10612 - Thanks so much for taking the time for the detailed reply. To be honest, I think (with the help of LLMs) I've gotten a little overconfident here. While I work in IT by trade, write a lot of fairly complex scripts, and have a deep understanding of networking, containerization, etc., I am absolutely _not_ a developer, and probably bit off a bit more than I can chew with this.

I really don't want to generate noise for you guys, I love the project and want to help out in any way I can, but moving forward I'll be more thoughtful about the scope of what I realistically should be submitting. Thanks again!

oschwartz10612:
No worries! Thanks for giving it a shot! :}

PR #3118, opened 2026-05-20
Fix public resource health with unknown WireGuard targets

### LINKED ISSUE #3105: Public ressource gets degraded if wireguard is added to the list of ressource sites
### Describe the Bug

I have a public resource with a multi-failover configuration: WireGuard, Newt, and Newt-Systemd. As soon as I add WireGuard to the public resource, performance degrades. This seems to be because the WireGuard site does not support health checks

Image

Image

### Environment

- Pangolin Version: 1.18.4 (CE)
- Gerbil Version: 1.4.0
- Traefik Version: 3.7.1
- Newt Version: 1.12.5

### To Reproduce

Add wireguard and newt site.
Add newt and wireguard to public ressource.
Visit Resources > Public page.
Ressource degrades.

### Expected Behavior

Health Column should reflect the health status of sites if they provide thiese checks and should stay healthy even if a wireguard site is added that does not support health checks.

Adityakk9031:
## Description

Fixes #3105.

Public resources were marked as degraded when a WireGuard target was added alongside healthy Newt targets. WireGuard targets do not report target health checks, so their `hcHealth` is stored as `unknown`.

The resource health aggregation previously treated any mixed state that was not all healthy, all unhealthy, or all unknown as degraded. That made `healthy + healthy + unknown` become `degraded`.

This change ignores `unknown` target health when at least one monitored target exists. Unknown targets remain neutral in the overall resource health calculation.

Expected behavior after this change:

- `healthy + healthy + unknown` -> `healthy`
- `healthy + unhealthy + unknown` -> `degraded`
- `unhealthy + unhealthy + unknown` -> `unhealthy`
- `unknown + unknown` -> `unknown`

### DIFF
--- a/server/lib/alerts/events/healthCheckEvents.ts
+++ b/server/lib/alerts/events/healthCheckEvents.ts
@@ -221,10 +221,18 @@ async function handleResource(
         )
         .where(eq(targets.resourceId, resource.resourceId));
 
+    const monitoredTargets = otherTargets.filter(
+        (t) => t.hcHealth !== "unknown"
+    );
+
     let health = "healthy";
-    const allUnknown = otherTargets.every((t) => t.hcHealth === "unknown");
-    const allHealthy = otherTargets.every((t) => t.hcHealth === "healthy");
-    const allUnhealthy = otherTargets.every((t) => t.hcHealth === "unhealthy");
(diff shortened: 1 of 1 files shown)


oschwartz10612 (approved):

oschwartz10612:
Thank you!

PR #3119, opened 2026-05-20
Sort resource filter options in audit logs

### LINKED ISSUE #3086: Wrong sort on ressource filter (http log) - maybe other locations as well?
### Describe the Bug

So i noticed, that the filterdropdown for resources in the http log is not ordered as the (public) ressource view.

Public ressource view is sorted by string character value (A = 65,0x41; a = 97,0x61). So the order of ` A, a, b, B, Test, test` would be `A, B, Test, a, b, test` - I personally would prefer a case insensitive sort order but its not my software.

Image

But the sorting in the filterdropdown looks like this:

Image

I assume its sorted by rowresult? Maybe Id of the ressource or by order of log appearance?

Wouldn't it be better, to have the same sorting for the same entity on all places where that entity is appears?

Yeah I'm nitpicking, sorry. 😅 

### Environment

- Pangolin Version: 1.18.4

### To Reproduce

Open http reqeust log, load data with the refresh icon, click on the filter icon on "Resource" to open the dropdown and have a non alphabetical order and not identical order to (public) resources.

### Expected Behavior

Identical sorting as on public resources.

Adityakk9031:
## Description

Fixes #3086.

The resource filter dropdown in the HTTP request log was using the order returned by the audit-log filter attributes API. That API collected distinct resource IDs from log rows, fetched resource names, and returned the merged result without sorting. As a result, the dropdown order could appear ID-based, log-order-based, or otherwise inconsistent with the public resources view.

This change sorts resource filter options by resource name before returning them from the audit log APIs. It also applies the same ordering to access and connection log resource filters so similar dropdowns stay consistent.

### DIFF
--- a/server/private/routers/auditLogs/queryAccessAuditLog.ts
+++ b/server/private/routers/auditLogs/queryAccessAuditLog.ts
@@ -93,6 +93,20 @@ export const queryAccessAuditLogsCombined = queryAccessAuditLogsQuery.merge(
 );
 type Q = z.infer<typeof queryAccessAuditLogsCombined>;
 
+function sortNamedFilterOptions<T extends { id: number; name: string | null }>(
+    items: T[]
+): T[] {
+    return [...items].sort((a, b) => {
+        const nameA = a.name ?? "";
+        const nameB = b.name ?? "";
+
+        if (nameA < nameB) return -1;
+        if (nameA > nameB) return 1;
+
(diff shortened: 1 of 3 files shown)


oschwartz10612:
Thank you! :}

oschwartz10612 (approved):

PR #3148, opened 2026-05-25
fix(audit-logs): route request audit log reads through logsDb

bishnubista:

## Description

Route the read paths in `server/routers/auditLogs/queryRequestAuditLog.ts` and `server/routers/auditLogs/queryRequestAnalytics.ts` through `logsDb` instead of `primaryLogsDb`, matching the existing private audit log routes (`queryActionAuditLog`, `queryAccessAuditLog`, `queryConnectionAuditLog` all already use `logsDb`).

In Postgres deployments configured with read replicas via `withReplicas` (`server/db/pg/logsDriver.ts`), this lets high-volume audit log reads use the replica pool instead of contending with writes on the primary. No-op on SQLite where `logsDb === primaryDb`.

### Why not also rewrite the facet queries?

While here, I benchmarked a candidate rewrite for the `// TODO: SOMEONE PLEASE OPTIMIZE THIS!!!!!` at `queryRequestAuditLog.ts:246`. Replacing the six parallel \`selectDistinct\` queries with a single \`UNION ALL\` over six \`GROUP BY ... LIMIT 500\` arms is **48-61% slower** on SQLite at 100k / 300k / 1M rows (20 runs per impl per size):

| Size | OLD \`SELECT DISTINCT LIMIT 501\` | \`UNION ALL\` rewrite | 6 × \`Promise.all\` \`GROUP BY\` |
|---|---:|---:|---:|
| 100k | 103.5 ms | 153.4 ms (+48.1%) | 153.6 ms (+48.3%) |
| 300k | 629.7 ms | 1015.1 ms (+61.2%) | 972.6 ms (+54.4%) |
| 1M   | 3152.1 ms | 4821.9 ms (+53.0%) | 4825.9 ms (+53.1%) |

\`EXPLAIN QUERY PLAN\` shows each grouped arm does \`USE TEMP B-TREE FOR GROUP BY\` — every distinct value is materialized before \`LIMIT\` truncates. The current DISTINCT+LIMIT short-circuits via hash dedup with early exit, which wins for the low-cardinality dropdown shape this endpoint produces. The long-term fix is likely a materialized facets table refreshed by a background job, not a query-shape rewrite.

## How to test?

1. \`npx tsc --noEmit\` — no new errors introduced in the changed files (same pre-existing error count before and after).
2. Manual check on an OSS SQLite stack: load \`/[orgId]/settings/logs/request\` in the dashboard; confirm the response shape (logs, pagination, filter dropdowns) is unchanged.
3. SaaS verification: in a Postgres deployment with \`postgres_logs.replicas\` configured, observe that requests to \`GET /org/{orgId}/logs/request\` route through the replica pool instead of incrementing query counters on the primary log DB.

### DIFF
--- a/server/routers/auditLogs/queryRequestAnalytics.ts
+++ b/server/routers/auditLogs/queryRequestAnalytics.ts
@@ -1,4 +1,4 @@
-import { logsDb, requestAuditLog, driver, primaryLogsDb } from "@server/db";
+import { logsDb, requestAuditLog, driver } from "@server/db";
 import { registry } from "@server/openApi";
 import { NextFunction } from "express";
 import { Request, Response } from "express";
@@ -74,12 +74,12 @@ async function query(query: Q) {
         );
     }
 
-    const [all] = await primaryLogsDb
+    const [all] = await logsDb
         .select({ total: count() })
         .from(requestAuditLog)
         .where(baseConditions);
 
-    const [blocked] = await primaryLogsDb
+    const [blocked] = await logsDb
         .select({ total: count() })
         .from(requestAuditLog)
         .where(and(baseConditions, eq(requestAuditLog.action, false)));
@@ -90,7 +90,7 @@ async function query(query: Q) {
 
     const DISTINCT_LIMIT = 500;
 
-    const requestsPerCountry = await primaryLogsDb
+    const requestsPerCountry = await logsDb
         .selectDistinct({
             code: requestAuditLog.location,
             count: totalQ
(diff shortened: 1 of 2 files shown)


oschwartz10612:
LGTM thanks for the simple fix! :} 

Also appreciate the write up. Agreed it seems like a future cache of the facets would be ideal. We could perhaps use redis for this.

oschwartz10612 (approved):

bishnubista:
> LGTM thanks for the simple fix! :}
> 
> Also appreciate the write up. Agreed it seems like a future cache of the facets would be ideal. We could perhaps use redis for this.

Thanks for the quick merge! Following up on the cache idea — I'm thinking of caching just the request audit log filter facets with a short TTL via the existing `#dynamic/lib/cache` abstraction, with the time range bucketed so dashboard refreshes actually hit. Does 60s or 300s feel right?

oschwartz10612:
Yeah sure I think this makes sense. Lets try 60?

bishnubista:
> Yeah sure I think this makes sense. Lets try 60?

Here [PR 3160](https://github.com/fosrl/pangolin/pull/3160)

PR #3151, opened 2026-05-26
Installer: Handle both Maxmind Country and ASN databases.

shleeable:

## Description
* This is expected to work but only the Country database was downloaded.
* trying to fix the left over folder.

## How to test?
As per documentation, https://docs.pangolin.net/self-host/advanced/enable-asn-lookup#install-with-the-installer

### DIFF
--- a/install/config/config.yml
+++ b/install/config/config.yml
@@ -22,7 +22,8 @@ server:
         methods: ["GET", "POST", "PUT", "DELETE", "PATCH"]
         allowed_headers: ["X-CSRF-Token", "Content-Type"]
         credentials: false
-    {{if .EnableGeoblocking}}maxmind_db_path: "./config/GeoLite2-Country.mmdb"{{end}}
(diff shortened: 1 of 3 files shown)


oschwartz10612:
Thanks for this! But instead of asking two questions could we just ask one? Like "do you want to download the maxmind asn and country databases for blocking"?

shleeable:
No problem. I'll merge them.. I doubt people would want one and not the other.

I was just giving people options.

shleeable:
> Thanks for this! But instead of asking two questions could we just ask one? Like "do you want to download the maxmind asn and country databases for blocking"?

I believe this is ready for review.

oschwartz10612 (approved):

oschwartz10612:
Thank you!

shleeable:
> Thank you!

I noticed this wasn't "linted" and has a few spaces... won't break anything but depending on your expectations might be worth running.

PR #3152, opened 2026-05-26
Installer: Bootstrap optional PostgreSQL/Redis

shleeable:

## Description
* Added postgres and redis to the docker-compose.yml using existing conditionals

### DIFF
--- a/install/config/config.yml
+++ b/install/config/config.yml
@@ -37,7 +37,7 @@ flags:
     disable_user_create_org: false
     allow_raw_resources: true
 
-{{if .IsAdvanced}}
+{{if .IsPostgreSQL}}
 postgres:
-  connection_string: postgresql://pangolin:{{.IsAdvancedPass}}@postgres:5432/pangolin
+  connection_string: postgresql://pangolin:{{.IsPostgreSQLPass}}@postgres:5432/pangolin
 {{end}}
--- a/install/config/config.yml
+++ b/install/config/config.yml
@@ -36,3 +36,8 @@ flags:
     disable_signup_without_invite: true
     disable_user_create_org: false
     allow_raw_resources: true
+
+{{if .IsAdvanced}}
+postgres:
+  connection_string: postgresql://pangolin:{{.IsAdvancedPass}}@postgres:5432/pangolin
+{{end}}
--- a/install/config/docker-compose.yml
+++ b/install/config/docker-compose.yml
@@ -1,7 +1,7 @@
 name: pangolin
 services:
   pangolin:
-    image: docker.io/fosrl/pangolin:{{if .IsEnterprise}}ee-postgres-{{end}}{{.PangolinVersion}}
+    image: docker.io/fosrl/pangolin:{{if .IsEnterprise}}ee-{{end}}{{if .IsAdvanced}}postgresql-{{end}}{{.PangolinVersion}}
     container_name: pangolin
     restart: unless-stopped
     deploy:
@@ -10,6 +10,16 @@ services:
           memory: 1g
         reservations:
           memory: 256m
+{{if .IsAdvanced}}
+    depends_on:
+      postgres:
+        condition: service_healthy
+      redis:
+        condition: service_healthy
+    networks:
+      - default
+      - backend
+{{end}}
     volumes:
       - ./config:/app/config
     healthcheck:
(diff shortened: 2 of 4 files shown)


shleeable:
Pangolin doesn't use .env as far as I am aware.

Is this a hard restriction? or can we create one for the ee condition to handle the passwords?

shleeable:
This is closer to something stable.

LaurenceJJones:
I like the thinking behind this, but I’m hesitant about making it the default.

My main concern is that it could make the Enterprise Edition more complicated for self-hosted users. We already hear from some users that the current component model can be difficult to maintain, so adding two more components by default feels like it could increase operational overhead without a clear benefit for most deployments.

Could you expand a bit more on what you see as the main advantage of this change?

To be upfront, I’m especially interested in why this should benefit the majority of users. My initial read is that the people who need HA may be happy to move to something like Postgres as part of that setup, but I’d expect most users are not running HA and may prefer to keep the default deployment as simple as possible.

shleeable:
I could put it behind another question? use sqllite or postgres etc.

My motivation is a bug I've experienced and spoke about in the discord.

image

I'm building a postgres/redis backed EE selfhosted instance now and going to try to migrate the mastodon again to this new pangolin.

If I experience the same issue with the latency burst, then it's not sqlites problem, it might be my routing but I've got 0.25ms pings on my traceroute.

sidenote: I've also had problems with hitting refresh on the "http request logs" and just killing the whole pangolin instance using sqlite (might be a memory issue but I'm not hitting my memory limit or OOMing).

shleeable:
nginx
image

pangolin on sqlite
image

site to pangolin ping
image

shleeable:
> I like the thinking behind this, but I’m hesitant about making it the default.

Is that more suitable? Giving people the choice to escalate into it?

AstralDestiny:
Well if you are wanting an HA you are going to selectively go for postgres and valkey. if it's a single instance then postgres isn't really being helpful there, Sure you can argue for it being better but postgres is bulky on memory. 

As for your latency would ask if it happens on more modern kernels or not and your instances are on vm's you control or shared hosting that might be overprovisioned? If your hosting the vm's are they type 1 vm's or type 2?

Anyways if you give me a few hours I'll deploy a mastadon and see if I encounter the same latency hike on a sqlite.

AstralDestiny:
Though if you want more control I'd point you at adding, https://github.com/orgs/fosrl/discussions/3080 to enable the other methods traefik can do as traefik by default does WRR, But there's better methods for doing HA across many backends.

oschwartz10612:
Yeah we are fine with a flag in the installer for EE or enterprise if you want to use postgres not sqlite. I would not think we need to put redis in there at all or prompt it. I think it would actually be slower and more overhead for most installs. Its really only required for HA installs which are currently not available self serve

shleeable:
So. I've just installed this setup on my host and I've noticed a MASSIVE improvement with the admin portal for starters.

The logs/analytics results are instant instead of taking multiple seconds.... and I had a bunch of problems with timeout errors.

shleeable:
> Well if you are wanting an HA you are going to selectively go for postgres and valkey. if it's a single instance then postgres isn't really being helpful there, Sure you can argue for it being better but postgres is bulky on memory.

So. I've just installed this setup on my host and I've noticed an improvement with the admin portal for starters.

The logs/analytics results are instant instead of taking multiple seconds.... and I haven't noticed any of problems with timeout errors I've experienced previously.

```
traefik  | 2026-05-26T02:51:07Z ERR Provider error, retrying in 748.949947ms error="cannot fetch configuration data: do fetch request: Get \"http://pangolin:3001/api/v1/traefik-config\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)"
gerbil    | INFO: 2026/05/25 20:16:26 Failed to report peer bandwidth: failed to send bandwidth data: Post "http://pangolin:3001/api/v1/gerbil/receive-bandwidth": read tcp 172.18.0.3:54690->172.18.0.2:3001: read: connection reset by peer
gerbil    | ERROR: 2026/05/26 02:51:24 Failed to notify server: Post "http://pangolin:3001/api/v1/gerbil/update-hole-punch": read tcp 172.18.0.3:60438->172.18.0.2:3001: read: connection reset by peer
gerbil    | ERROR: 2026/05/26 02:51:49 Failed to notify server: Post "http://pangolin:3001/api/v1/gerbil/update-hole-punch": dial tcp 172.18.0.2:3001: i/o timeout
```

> As for your latency would ask if it happens on more modern kernels or not and your instances are on vm's you control or shared hosting that might be overprovisioned? If your hosting the vm's are they type 1 vm's or type 2?

These are all dedicated... and the box I tested the regression is `Ubuntu 26.04 LTS (GNU/Linux 7.0.0-15-generic`.
Proxmox (qemu) based on EYPC Zen3.

image

shleeable:
> Yeah we are fine with a flag in the installer for EE or enterprise if you want to use postgres not sqlite. I would not think we need to put redis in there at all or prompt it. I think it would actually be slower and more overhead for most installs. Its really only required for HA installs which are currently not available self serve

I've noticed there are both `ee-postgresql` and just `postgres` images? I assume that means the postgres image is available for the standard/community edition? 

I'll move the question out of the enterprise path and ask which DB regardless.

Redis is an excellent "does one job" tool... how are you handling internal data caching for things that would be used by redis? 

(sidenote: I'm running redis on that pangolin deployment above and I've never seen it actually record any records to redis... so I don't even know if it works)

AstralDestiny:
https://github.com/user-attachments/assets/729f78c2-0f04-49d3-97b1-78ef821d8b7c

Is it supposed to be faster then this? Not on the same network for me.

shleeable:
> pangolinee.mp4
> 
> Is it supposed to be faster then this? Not on the same network for me.

My server with sqlite takes... 5-10 seconds to load those logs.

Snappy like your video after I moved to postgres/redis.

https://github.com/user-attachments/assets/3300f545-5b94-42c2-87e5-77c8be5b4f2e

shleeable:
Please squash when you merge this. :)

I think it's ready for review... unless you absolutely want to pull out redis

AstralDestiny:
Yeah I'm on sqlite for that video.

shleeable:
I am not anti-sqlite.. there is obviously a problem that's unrelated to this PR.

I'll try to bootstrap another dev VM using sqlite again and see if I can replicate my problems... but these servers are modern on NVME storage so It's not the kind of behaviour I'd expect from sqlite.

AstralDestiny:
Should also mention that host is Debian 13

AstralDestiny:
No no just trying to figure it out as I've seen others mention latency issues and whatnot so trying to compare what you have and what they have. See if there's anything that matches or not. I'm all for this being an option.

oschwartz10612:
Hey thanks for this. I was about to take it but then I realized that redis is only available in the enterprise edition but it looks like it always asks the question. You could split it into two questions: one for postgres at no matter what the enterprise answer is and the other for redis if they are on the ee.

shleeable:
> Hey thanks for this. I was about to take it but then I realized that redis is only available in the enterprise edition but it looks like it always asks the question. You could split it into two questions: one for postgres at no matter what the enterprise answer is and the other for redis if they are on the ee.

Easy.

shleeable:
@oschwartz10612 Ok. That added a small amount of logic to the docker compose.... but looks a little nicer overall.

oschwartz10612:
LGTM thanks for making the change! :}

oschwartz10612 (approved):

PR #3154, opened 2026-05-26
fix(logs): refresh default end time

RitwijParmar:
## Summary
- advances the log query end time to the current time when refreshing the default live range
- preserves an explicit end time when the user has selected one in the URL
- applies the behavior consistently across request, access, action, and connection log pages

## Why
The log pages initialize their default end time to now, but the refresh action reused that stale timestamp. For a user watching logs with the default range, refresh could keep querying the old upper bound instead of including newly written rows.

## Verification
- git diff --check
- attempted npm run dev:check, but this checkout has no node_modules installed and npx tried to install the unrelated tsc package instead of using a local TypeScript compiler

### DIFF
--- a/src/app/[orgId]/settings/logs/access/page.tsx
+++ b/src/app/[orgId]/settings/logs/access/page.tsx
@@ -280,10 +280,14 @@ export default function GeneralPage() {
         console.log("Data refreshed");
         setIsRefreshing(true);
         try {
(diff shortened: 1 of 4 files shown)


oschwartz10612 (approved):

oschwartz10612:
Thank you!

PR #3160, opened 2026-05-26
feat(audit-logs): cache request audit log filter facets

bishnubista:

## Description

Followup to #3148. As discussed in that thread, this adds a cache-aside to `queryUniqueFilterAttributes` so that dashboard refreshes within the same minute don't re-run the six `selectDistinct` queries that populate the audit log filter dropdowns.

Implementation:
- **Import**: `regionalCache as cache` from `#dynamic/lib/cache`, matching the dynamic pattern from d1fb2e19.
- **Cache key**: `cache:audit-log-facets:v1:{orgId}:{timeStartBucket}:{timeEndBucket}`, with timestamps bucketed to 60s boundaries so refresh-within-the-minute hits cache. The `v1` segment guards against future shape changes.
- **TTL**: 60s, per "Lets try 60?" in #3148.
- **Failure isolation**: get/set errors are caught and logged via `logger.warn`; the request never 500s on cache failure.
- **Combined effect**: facet dropdowns become approximate to within ~60s. The six underlying queries are unchanged - the prior bench in #3148 showed their current `SELECT DISTINCT LIMIT 501` shape already short-circuits via hash dedup with early exit, so a query rewrite would have regressed.

Branch note: this PR is based on `dev`, so the file still references `primaryLogsDb` rather than `logsDb`. #3148's routing fix landed on `main`; the changes are independent and will compose cleanly when the branches sync.

## How to test?

1. `npx tsc --noEmit` — no new errors in the changed file (22 pre-existing errors counted before and after).
2. Manual smoke on an OSS-SQLite stack: load `/[orgId]/settings/logs/request` in the dashboard, then refresh within 60 seconds. Confirm filter dropdowns populate identically on both requests and that DB query counters don't double. Avoid testing across minute boundaries — bucket rollover would look like a miss.
3. SaaS verification (optional): with `enable_redis: true`, confirm `cache.getCurrentBackend()` returns `"redis"` for the facet keys.

### DIFF
--- a/server/routers/auditLogs/queryRequestAuditLog.ts
+++ b/server/routers/auditLogs/queryRequestAuditLog.ts
@@ -12,6 +12,7 @@ import { QueryRequestAuditLogResponse } from "@server/routers/auditLogs/types";
 import response from "@server/lib/response";
 import logger from "@server/logger";
 import { getSevenDaysAgo } from "@app/lib/getSevenDaysAgo";
+import { regionalCache as cache } from "#dynamic/lib/cache";
 
 export const queryAccessAuditLogsQuery = z.object({
     // iso string just validate its a parseable date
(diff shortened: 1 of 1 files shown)


bishnubista:
@oschwartz10612 Any thoughts?

PR #3169, opened 2026-05-27
Add in-app documentation shortcuts

malsomesh9:
## Summary

This draft PR starts addressing #2724 by adding clearer documentation entry points inside the app.

## What I changed

- add a persistent documentation shortcut in the sidebar footer so users can reach the docs from anywhere in the app
- add a documentation shortcut inside the support popover, including the fallback state when support requests are unavailable
- keep the change limited to shared UI components so it improves discoverability without introducing page-specific doc mappings yet

## Plan

If this direction looks good, the next follow-up would be to add page-specific doc links for the highest-friction settings flows rather than scattering generic help text everywhere at once.

## Validation

- ran `npm exec --yes prettier -- --check src/components/LayoutSidebar.tsx src/components/SidebarSupportButton.tsx`

### DIFF
--- a/src/components/LayoutSidebar.tsx
+++ b/src/components/LayoutSidebar.tsx
@@ -19,7 +19,13 @@ import { approvalQueries } from "@app/lib/queries";
 import { build } from "@server/build";
 import { useQuery } from "@tanstack/react-query";
 import { ListUserOrgsResponse } from "@server/routers/org";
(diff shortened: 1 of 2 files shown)


oschwartz10612:
Could you provide some screenshots of how this looks - interested in exploring the direction!

PR #3188, opened 2026-05-29
Add global Ctrl+K command palette for navigation and search.

LaurenceJJones:

## Description

Provides quick access to sidebar destinations, org switching, server-filtered entity lookup, and common actions from anywhere in the authenticated layout.

## How to test?

When loading any pages you can access a global command palette which allows for navigation using keyboard only as well as searching possibilities. Opening as draft for now whilst we refine this implementation.


AstralDestiny:
Ooh yes please.

PR #3251, opened 2026-06-12
Fix unescaped regex metacharacters in PATH rule matching causing request failures

kshitijshresth:

## Summary

`isPathAllowed` (used by `checkRules` in `verifySession.ts` to evaluate PATH resource rules) converts wildcard pattern segments to regular expressions without escaping regex metacharacters. `isValidUrlGlobPattern` in `validators.ts` explicitly permits `- . _ ~ ! $ & ' ( ) * + , ; # = @ :` in rule values, so rules that pass API validation can produce invalid or semantically wrong regexes at request time.

## Impact

**Unhandled exception on every request.** A PATH rule value like `/(api*` is accepted by validation, but the matcher builds `new RegExp("^(api.*$")`, which throws `SyntaxError: Invalid regular expression: missing )`. `checkRules` has no exception handling, so the error propagates to the catch-all in `verifyResourceSession` and every request to that resource fails with `500 Failed to verify session` until the rule is deleted.

**Incorrect matching for literal characters.** Since PATH rules commonly gate auth bypass (ACCEPT) or denial (DROP), both directions matter:

* `*.png` matched `imageXpng` (`.` acted as regex any-char)
* `a+b*` matched `aaabc` and failed to match `a+bc` (`+` acted as a quantifier)
* `$ref*` could never match anything (`$` acted as an anchor)

## Changes

* Extracted `isPathAllowed` from `verifySession.ts` into `pathMatch.ts` as a pure module (no logger/config/db imports) so it can be unit tested in isolation; `verifySession.ts` imports and re-exports it, keeping its public surface unchanged.
* Regex metacharacters are now escaped before the `*` → `.*` and `?` → `.` substitutions. The escape set deliberately excludes `*` and `?` so wildcard semantics are unchanged.
* Compiled segment regexes are cached in a module-level `Map`. Keys are admin-defined rule pattern segments (a bounded set), not request paths, so the cache cannot grow with traffic. This avoids recompiling regexes on every proxied request, which sits on the hot path.
* As part of the extraction, the per-recursion-step `logger.debug` calls (~10 string interpolations per segment comparison, evaluated eagerly on every request regardless of log level) were removed.

## Tests

`verifySession.test.ts` previously tested a stale inline duplicate of `isPathAllowed` rather than the real implementation, so regressions in the production code could never fail the suite. It now imports the real function from `pathMatch.ts`.

* All pre-existing assertions are unchanged and pass, demonstrating no behavioral change for patterns without metacharacters.
* Added assertions covering `( ) [ ] { } | \ . + ^ $ |` as literals, a no-throw guarantee for all characters accepted by `isValidUrlGlobPattern`, preserved `?` single-character wildcard behavior, and a deep-path sanity check.

Verified against the unfixed logic first: the new tests fail with `Invalid regular expression: /^(api.*$/: Unterminated group`, and pass after the fix.

```bash
npx tsx server/routers/badger/verifySession.test.ts   # all tests pass
npx tsc --noEmit                # clean
npx prettier --check <changed files>   # clean
```

## Reproduction (before this fix)

1. Create a resource with rules enabled.
2. Add a PATH rule with value `/(api*` (accepted by validation).
3. Every request to the resource now returns 500 (`Invalid regular expression: /^(api.*$/: Unterminated group` in logs).

### DIFF
--- a/server/lib/pathMatch.ts
+++ b/server/lib/pathMatch.ts
@@ -1,0 +1,74 @@
+const MAX_RECURSION_DEPTH = 100;
+
+const segmentRegexCache = new Map<string, RegExp>();
+
+function getSegmentRegex(patternPart: string): RegExp {
+    let regex = segmentRegexCache.get(patternPart);
+    if (!regex) {
+        const regexPattern = patternPart
+            .replace(/[.+^${}()|[\]\\]/g, "\\$&")
+            .replace(/\*/g, ".*")
+            .replace(/\?/g, ".");
+        regex = new RegExp(`^${regexPattern}$`);
+        segmentRegexCache.set(patternPart, regex);
+    }
+    return regex;
+}
+
+export function isPathAllowed(pattern: string, path: string): boolean {
+    const normalize = (p: string) => p.split("/").filter(Boolean);
+    const patternParts = normalize(pattern);
+    const pathParts = normalize(path);
+
+    function matchSegments(
+        patternIndex: number,
+        pathIndex: number,
+        depth: number = 0
+    ): boolean {
+        if (depth > MAX_RECURSION_DEPTH) {
+            return false;
+        }
+
+        const currentPatternPart = patternParts[patternIndex];
+        const currentPathPart = pathParts[pathIndex];
+
+        if (patternIndex >= patternParts.length) {
+            return pathIndex >= pathParts.length;
+        }
+
+        if (pathIndex >= pathParts.length) {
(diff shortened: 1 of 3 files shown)


oschwartz10612 (approved):

oschwartz10612:
Thank you! 

Would you also be able to bring this change to the https://github.com/fosrl/pangolin-node repo? This is for the remote nodes from the cloud and would also experience this issue.

PR #3278, opened 2026-06-14
fix(badger): inject Remote-Groups header for downstream RBAC

rinseaid:

### Preface

This PR was developed with AI assistance. I drove the direction and tested everything, but the AI did the actual coding. I reviewed, tested, and validated all changes.

## Description

This PR addresses a missing RBAC feature in the Badger middleware.

Currently, `Remote-Groups` isn't passed to the proxy. I added `groups` to `BasicUserData` and mapped the user's roles to a comma-separated string in `isUserAllowedToAccessResource`. Badger now natively injects this header for downstream apps to use for RBAC, mirroring how `Remote-Role` is already handled.

## How to test?

Log in, access a protected resource, and check the downstream headers. `Remote-Groups` should be populated with the user's roles.

### DIFF
--- a/server/routers/badger/verifySession.ts
+++ b/server/routers/badger/verifySession.ts
@@ -71,6 +71,7 @@ type BasicUserData = {
     email: string | null;
     name: string | null;
     role: string | null;
+    groups?: string;
 };
 
 export type VerifyUserResponse = {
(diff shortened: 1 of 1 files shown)


oschwartz10612:
Hi! Thanks for the contribution. Why would we want both `Remote-Groups` and `Remote-Roles` both with the same values sent from badger? Is there something I am missing there? I think we would prefer to stick with just roles as that is what they are called across the platform.

rinseaid:
@oschwartz10612 the goal was compatibility with downstream apps that consume the remote groups header, but I realize now that most (if not all!) support changing which header they use. Thanks for replying, closing this out.

PR #1884, opened 2025-11-18
Bump Traefik to v3.6 due to Docker 29.X.X compatibility

v1rusnl:
## Community Contribution License Agreement
By creating this pull request, I grant the project maintainers an unlimited,
perpetual license to use, modify, and redistribute these contributions under any terms they
choose, including both the AGPLv3 and the Fossorial Commercial license terms. I
represent that I have the right to grant this license for all contributed content.

## Description
When upgrading to Docker 29.X.X, Traefik stops working because of Docker API version bump to min. 1.44 (before: 1.24), see https://github.com/traefik/traefik/issues/12253 . https://github.com/traefik/traefik/pull/12256 fixes this in Traefik v3.6.1. 

The version bump fixes a non working installation or upgrade when using Crowdsec since the Crowdsec compose part uses a Docker Label which does not work in Traefik v3.5.X after the Docker upgrade to 29.X.X.

## How to test?
Update Docker to 29.X.X with Traefik 3.5.X and Crowdsec enabled -> Pangolin and Resources do not work
Traefik Error: ```ERROR: CrowdsecBouncerTraefikPlugin: YYYY/MM/DD HH:MM:SS appsecQuery:unreachable```

Upgrade to Traefik 3.6 -> Pangolin and Resources are working

PR #3386, opened 2026-07-03
Upgrade Traefik image to version 3.7

v1rusnl:

## Description
Upgrade Traefik image to version 3.7

### DIFF
--- a/compose.example.yaml
+++ b/compose.example.yaml
@@ -41,7 +41,7 @@ services:
       - 80:80 # Port for traefik because of the network_mode
 
   traefik:
-    image: traefik:v3.6
+    image: traefik:v3.7
     container_name: traefik
     restart: unless-stopped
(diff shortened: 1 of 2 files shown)


AstralDestiny:
Looks good to me though If I recall there's one other location it needs to be bumped in.

v1rusnl:
> Looks good to me though If I recall there's one other location it needs to be bumped in.

Had that also in mind from the past but these where the only places with a reference to traefik:v3.6.

Was the same in #1884 now that I looked after it.

AstralDestiny:
I only say as I orginally had updated the traefik_dynamic in the past but nobody ever got the updated config in the end yet I had made the change, something related to the go files if I recall

oschwartz10612:
LGTM - will be in the next release

oschwartz10612 (approved):

PR #3410, opened 2026-07-08
fix: redirect to /auth/initial-setup after hitting auth rate limit (#…

### LINKED ISSUE #3408: Redir to /auth/initial-setup after run in login rate limit
### Describe the Bug

Pangolin landing page redirects to /auth/initial-setup (ERROR 1682261561) after running into the login rate limit when the limit is configured like

`rate_limits:
  auth:
    max_requests: 3
`

[log.txt](https://github.com/user-attachments/files/29791479/log.txt)

### Environment

- Ubuntu 26 LTS:
- Pangolin Version: 1.19.4
- Enterprise
- Gerbil Version: 1.4.2
- Traefik Version: 3.6
- Newt Version: no newt connection
- Client Version: Safari, Chrome

### To Reproduce

- Configure rate limit
`rate_limits:
  auth:
    max_requests: 3
`
- Open pangolin landing page: pangolin.domain.org
- Login with invalid credentials for 3 times
- See the rate limit notification
- Reload the page several times

### Expected Behavior

- See the login page with rate limit notification

Adityakk9031:
## Summary

Fixes #3408

After hitting the `rate_limits.auth` limit, reloading the dashboard root
redirected users to `/auth/initial-setup` instead of staying on the login
page — even on a fully configured instance.

## Root Cause

Two issues combined to cause the redirect:

**1. `GET /auth/initial-setup-complete` was inside the global `authRouter`
rate limiter scope.**

The `authRouter` applies a blanket rate limiter to every `/auth/*` route using
an `<IP>:<path>` key. The `/initial-setup-complete` endpoint is called on
**every dashboard root page load** (it's a pure boolean read with no secrets),
so with a low `max_requests` value it exhausted its own per-path budget after
just a few reloads and started returning `429`.

**2. The `page.tsx` root page silently swallowed the 429.**

```ts
// Before
let complete = false;
try {
    const res = await internal.get(`/auth/initial-setup-complete`, ...);
    complete = res.data.data.complete;
} catch (e) {}          // 429 discarded here
if (!complete) {        // complete is still false (default)
    redirect("/auth/initial-setup");  // fires incorrectly
}

### DIFF
--- a/server/routers/external.ts
+++ b/server/routers/external.ts
@@ -1338,6 +1338,12 @@ authenticated.get("/ws/round-trip-message/:messageId", checkRoundTripMessage);
 // Auth routes
 export const authRouter = Router();
 unauthenticated.use("/auth", authRouter);
+
+// Register setup-check BEFORE the global auth rate limiter.
+// This endpoint is called on every dashboard root page load (pure boolean
+// read, no secrets) and must not consume the auth rate-limit budget.
+authRouter.get("/initial-setup-complete", auth.initialSetupComplete);
+
 authRouter.use(
     rateLimit({
         windowMs:
@@ -1642,7 +1648,6 @@ authRouter.post("/idp/:idpId/oidc/generate-url", idp.generateOidcUrl);
 authRouter.post("/idp/:idpId/oidc/validate-callback", idp.validateOidcCallback);
 
 authRouter.put("/set-server-admin", auth.setServerAdmin);
-authRouter.get("/initial-setup-complete", auth.initialSetupComplete);
 authRouter.post("/validate-setup-token", auth.validateSetupToken);
 
 // Security Key routes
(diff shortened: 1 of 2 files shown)


AstralDestiny:
Fixes something couldn't replicate though?

Sorry if I come off terse or whatnot, Not trying to be, Just wasn't able to replicate the issue they mentioned with their steps on EE, So the question is what did you do differently to replicate or just went to the code and tried to find the reference for that? or weren't able to replicate either but went to find a fix?

Adityakk9031:
Hi @AstralDestiny, 
No worries at all, it's a completely valid question! The reason this issue is a bit tricky to replicate lies in how Next.js server-side fetches, header forwarding, and the path-based rate limiter interact.

Adityakk9031:
@miloschwartz have a look please

Adityakk9031:
@oschwartz10612 have a look

miloschwartz (approved):

PR #3459, opened 2026-07-19
Fix request audit log ordering

### LINKED ISSUE #3458: HTTP Request logs mix up data in rows
### Describe the Bug

When viewing the http request logs, I'm getting a lot of allowed from random IPs to services which are 100% blocked from these IP addresses. 

When the logs are exported to CSV, I can see the correct data and everything lines up; So it's just a mix up in the GUI somewhere. 

Image

Image

### Environment

- OS Type & Version: Debian 12
- Pangolin Version: v1.20.0-ee
- Edition (Community or Enterprise):
- Gerbil Version: 1.4.2
- Traefik Version: 3.7
- Newt Version: 1.14
- Client Version: N/A

### To Reproduce

Any activity and check the logs 

(I have a high amount of activity because Homepage dashboard triggers a heap of activity with the widgets)

### Expected Behavior

GUI to match what is exported in the CSV. 

As of currently, it's impossible to use the GUI to diagnose anything

shubhamsinnh:
## Summary

- Add a deterministic secondary sort for request audit logs when multiple entries share the same timestamp.
- Stop rendering generated sample rows in audit log tables while real log data is loading.

## Why

Request audit log entries store timestamps at second-level precision. Under high activity, many entries can share the same timestamp. The request log query sorted only by timestamp, so separate executions for the GUI and CSV export could return same-timestamp rows in different orders. That can make row-by-row comparisons look like the GUI mixed data even when each individual row is valid.

Request logs now sort by timestamp descending and then id descending, matching the deterministic ordering pattern already used by action, access, and connection audit log queries.

Audit log pages also now render only real API data instead of generated placeholder rows while loading.

Fixes #3458.

## Validation

- `git diff --check HEAD~1 HEAD`
- `npx prettier --check server/routers/auditLogs/queryRequestAuditLog.ts 'src/app/[orgId]/settings/logs/access/page.tsx' 'src/app/[orgId]/settings/logs/action/page.tsx' 'src/app/[orgId]/settings/logs/connection/page.tsx' 'src/app/[orgId]/settings/logs/request/page.tsx'`
- `npx tsc --noEmit --pretty false`
- Inserted a local burst of 50 same-timestamp request audit rows to verify the high-activity shape, then removed the synthetic rows afterward.

### DIFF
--- a/server/routers/auditLogs/queryRequestAuditLog.ts
+++ b/server/routers/auditLogs/queryRequestAuditLog.ts
@@ -158,7 +158,7 @@ export function queryRequest(data: Q) {
         })
         .from(requestAuditLog)
         .where(getWhere(data))
-        .orderBy(desc(requestAuditLog.timestamp));
+        .orderBy(desc(requestAuditLog.timestamp), desc(requestAuditLog.id));
 }
 
 async function enrichWithResourceDetails(
--- a/src/app/[orgId]/settings/logs/access/page.tsx
+++ b/src/app/[orgId]/settings/logs/access/page.tsx
@@ -23,7 +23,6 @@ import { usePaidStatus } from "@app/hooks/usePaidStatus";
 import { tierMatrix } from "@server/lib/billing/tierMatrix";
 import { logQueries } from "@app/lib/queries";
 import { useQuery } from "@tanstack/react-query";
-import type { QueryAccessAuditLogResponse } from "@server/routers/auditLogs/types";
 
 export default function GeneralPage() {
     const router = useRouter();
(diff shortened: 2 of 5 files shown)


shubhamsinnh:
Hi @oschwartz10612 , @miloschwartz,
Have opened the PR to fix the request audit log ordering issue.

Would you mind taking a look? happy to make any changes if required.

shubhamsinnh:
Hi @oschwartz10612 @miloschwartz - a quick follow-up on the request audit log ordering fix. The change adds deterministic ordering for same-timestamp entries and removes synthetic loading rows; please let me know if you would like any adjustments or additional validation before review.

shubhamsinnh:
Hi @oschwartz10612 @miloschwartz  just checking whether there’s any update on this PR when you have a chance. It is still awaiting review, and I’m happy to make any changes or provide additional validation if needed.

@AstralDestiny, would appreciate an update at your end as well.

miloschwartz:
Thanks for contributing. We don't want to remove the sample logs.

PR #3460, opened 2026-07-19
Fix Punycode domain validation

shubhamsinnh:
## Summary

- Allow Punycode TLDs to pass backend domain validation when they are present in the existing TLD allowlist.
- Add regression coverage for the reported IDN domain converted to Punycode.

## Root cause

The domain creation form converts internationalized domains to Punycode before submitting them. The backend validator accepted Punycode-looking labels, but rejected the final TLD unless it contained only letters. That blocked valid TLDs such as xn--p1ai before the existing allowlist check could accept them.

## Impact

Admins can create valid IDN domains after the UI converts them to Punycode. Unknown TLDs are still rejected through the existing allowlist.

Fixes #3427

## Validation

- `/tmp/node-v22.13.1-linux-x64/bin/node ./node_modules/tsx/dist/cli.mjs server/lib/validators.test.ts`
- `/tmp/node-v22.13.1-linux-x64/bin/node ./node_modules/prettier/bin/prettier.cjs --check server/lib/validators.ts server/lib/validators.test.ts`
- `/tmp/node-v22.13.1-linux-x64/bin/node ./node_modules/typescript/bin/tsc --noEmit --pretty false`
- `git diff --check`

PR #3461, opened 2026-07-19
Fix resource rule value validation

### LINKED ISSUE #3462: Resource rule APIs accept invalid country and ASN values
﻿### Describe the Bug

The resource rule create/update APIs do not consistently validate rule values for every supported match type.

`server/lib/validators.ts` has shared validation for `CIDR`, `IP`, `PATH`, `REGION`, `COUNTRY`, `COUNTRY_IS_NOT`, and `ASN`, and the UI-side policy access rule validation applies the same country/ASN checks. However, the backend resource rule create/update routes only validate `CIDR`, `IP`, `PATH`, and `REGION` values before writing the rule.

That means a direct API caller can create or update a resource rule with invalid `COUNTRY`, `COUNTRY_IS_NOT`, or `ASN` values that the shared validator would reject.

### Evidence

- `getResourceRuleValueValidationError` validates country and ASN values in `server/lib/validators.ts`.
- `createResourceRule` currently only validates CIDR/IP/PATH/REGION values before insert.
- `updateResourceRule` currently only validates CIDR/IP/PATH/REGION values before update.

### Expected Behavior

Resource rule create/update routes should use the same rule match list and value validation as the shared validator so direct API calls cannot persist invalid country or ASN rule values.

### Proposed Fix

Reuse `RESOURCE_RULE_MATCH_TYPES` and `getResourceRuleValueValidationError` in the create/update routes instead of maintaining a partial duplicate validation switch.

shubhamsinnh:
﻿## Community Contribution License Agreement

## Description

Fixes #3462.

Resource rule creation and update were using hand-rolled value validation that only covered `CIDR`, `IP`, `PATH`, and `REGION` rules. The shared rule validator also covers `COUNTRY`, `COUNTRY_IS_NOT`, and `ASN`, so direct API calls could persist invalid country or ASN values that the shared validation logic would reject.

This reuses the shared rule match list and value validator in both resource rule routes so backend API validation stays aligned with the existing validator behavior.

This also lets the create route accept `COUNTRY_IS_NOT`, matching the update route and the shared resource rule match list.

## Validation

- `node node_modules/prettier/bin/prettier.cjs --check server/routers/resource/createResourceRule.ts server/routers/resource/updateResourceRule.ts`
- `node node_modules/tsx/dist/cli.mjs server/lib/validators.test.ts`
- `git diff --check upstream/main...HEAD`

Note: full project TypeScript validation was attempted locally but did not complete before timeout; no diagnostics were emitted before the timeout.

### DIFF
--- a/server/routers/resource/createResourceRule.ts
+++ b/server/routers/resource/createResourceRule.ts
@@ -9,16 +9,14 @@ import createHttpError from "http-errors";
 import logger from "@server/logger";
 import { fromError } from "zod-validation-error";
 import {
-    isValidCIDR,
-    isValidIP,
-    isValidUrlGlobPattern
+    getResourceRuleValueValidationError,
+    RESOURCE_RULE_MATCH_TYPES
 } from "@server/lib/validators";
 import { OpenAPITags, registry } from "@server/openApi";
-import { isValidRegionId } from "@server/db/regions";
 
 const createResourceRuleSchema = z.strictObject({
     action: z.enum(["ACCEPT", "DROP", "PASS"]),
-    match: z.enum(["CIDR", "IP", "PATH", "COUNTRY", "ASN", "REGION"]),
+    match: z.enum(RESOURCE_RULE_MATCH_TYPES),
     value: z.string().min(1),
     priority: z.int(),
     enabled: z.boolean().optional()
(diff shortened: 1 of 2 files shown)


oschwartz10612:
Whoops! Sorry did not see this and fixed your bug report in ef13cbe7c9cb09d245b40ea82970dcee018f2d90. Apologies but thanks for the report and work!

shubhamsinnh:
> Whoops! Sorry did not see this and fixed your bug report in [ef13cbe](https://github.com/fosrl/pangolin/commit/ef13cbe7c9cb09d245b40ea82970dcee018f2d90). Apologies but thanks for the report and work!

Thanks for the update @oschwartz10612
 I also opened two other fixes:
- #3459 request audit log ordering
- #3460 Punycode domain validation

I would really appreciate if you could take a moment to review them.

PR #3510, opened 2026-07-30
Fix integration API site resource lookup

### LINKED ISSUE #2743: [Integration API] GET /site-resource/{siteResourceId} not working
### Describe the Bug

As said in the title this route in the integration API is not working.
Here is the error I get with a 400 status :
`'Validation error: Invalid input: expected string, received undefined at "siteId"; Invalid input: expected string, received undefined at "orgId"'`
After taking a quick look at the code (even tho I have not done any ts and zod for a looooong time), I believe this bug is due to [the zod schema](https://github.com/fosrl/pangolin/blob/main/server/routers/siteResource/getSiteResource.ts#L13) expecting a siteId and an orgId but [the route](https://github.com/fosrl/pangolin/blob/main/server/routers/external.ts#L312) does not contains it.
So either the getSiteResource function should be working without it either the route needs to be updated with the correct path params.

### Environment

I have been able to reproduce this bug in :
* free tier pangolin.net hosted
* self hosted v1.15.3 on NixOS

### To Reproduce

Just create a site-resource, get its id via `/org/{orgId}/site-resources` and then try to get it via `/site-resource/{siteResourceId}`. I think this issue is very old because it is happening on my selfhosted v1.15.3 AND on the app.pangolin.net hosted free tier.

### Expected Behavior

Should return me a site-resource.

shubhamsinnh:

## Description

Fixes #2743.

The integration API route `GET /site-resource/{siteResourceId}` only provides a resource ID, but the shared request schema required an `orgId`. As a result, valid requests failed validation before the resource lookup ran.

This makes `orgId` optional for ID-based lookups while keeping it as an additional constraint for org-scoped routes. It also updates the OpenAPI definitions so they only advertise parameters that are actually present in the route.

A regression test covers the ID-only route, the existing org-scoped form, and invalid resource IDs.

For API-key and user routes, the existing access middleware still authorizes the requested resource before this handler runs. Root API keys continue through the existing root-key check and receive the normal not-found response when the resource does not exist.

## How to test?

```bash
npx tsx server/routers/siteResource/getSiteResource.test.ts
npx eslint server/routers/siteResource/getSiteResource.ts server/routers/siteResource/getSiteResource.test.ts
npx prettier --check server/routers/siteResource/getSiteResource.ts server/routers/siteResource/getSiteResource.test.ts
npx tsc --noEmit
```

All four checks pass locally.

### DIFF
--- a/server/routers/siteResource/getSiteResource.test.ts
+++ b/server/routers/siteResource/getSiteResource.test.ts
@@ -1,0 +1,58 @@
+import { assertEquals } from "@test/assert";
+import { getSiteResourceParamsSchema } from "./getSiteResource";
+
+function testSiteResourceIdOnlyParams() {
+    const result = getSiteResourceParamsSchema.safeParse({
+        siteResourceId: "42"
+    });
+
+    assertEquals(
+        result.success,
+        true,
+        "siteResourceId-only integration routes should pass validation"
+    );
+
+    if (result.success) {
+        assertEquals(
+            result.data.siteResourceId,
+            42,
+            "siteResourceId should be parsed as a number"
+        );
+        assertEquals(
+            result.data.orgId,
+            undefined,
+            "orgId should remain optional"
+        );
+    }
+}
+
+function testOrgScopedParamsRemainSupported() {
+    const result = getSiteResourceParamsSchema.safeParse({
+        siteResourceId: "42",
+        orgId: "org-id"
+    });
+
+    assertEquals(
+        result.success,
+        true,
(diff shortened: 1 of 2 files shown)


shubhamsinnh:
Hi @oschwartz10612, @miloschwartz, and @AstralDestiny 
This addresses the integration API lookup failure reported in #2743.

I'd appreciate your review when you get a chance
Thanks !!

oschwartz10612 (approved):

oschwartz10612:
Thanks!

PR #3512, opened 2026-07-30
Fix: resolve batch certificates for NS & Wildcard domains 

Fredkiss3:

## Summary 

- Introduced in #3469 : 
   My logic to get certificates for NS & wildcard certificates was flawed, I first get the certificates for exact domains, then I check if there are domains with missing certificates (wildcard & NS certificates are not matched on the exact name), then look if there are matching wildcard certificates  or NS certificates for the domains left. The flaw was that, with my logic, the domains with missing certs would always be empty. 

## How to test?

Go to resources > public, all certificates should appear at once.


oschwartz10612 (approved):

oschwartz10612:
:raised_hands:

PR #3551, opened 2026-08-08
fix: add Traefik routing rules for Integration API (/v1) endpoints (#3334)

### LINKED ISSUE #3334: API is enabled yet all /v1 API endpoints return a 404 error.
### Describe the Bug

When I try to visit an endpoint in the API for example /v1/docs it will send me to a 404 error. The same thing happens for any endpoint in Pangolin. This has broken my services and I cannot do anything through the API.

I verified the files in the CONFIG and they all comply with the Docs shown in Pangolin.

### Environment

- OS Type & Version: Docker Compose + Ubuntu 24.
- Pangolin Version: 1.19.2
- Edition (Community or Enterprise): Enterprise
- Gerbil Version: [v0.18.2 - Fianchetto]
- Traefik Version: [v3.7.5]

### To Reproduce

Visit any endpoint at the api.tunnel.domain.tld at /v1/ adn it will always return error 404.

### Expected Behavior

200 OK
API Payload

Adityakk9031:
Fixes #3334 where enabling the Integration API (\enable_integration_api: true\) caused all \/v1/*\ endpoints (e.g. \/v1/docs\, \/v1/openapi.json\, \/v1/org/...\) to return 404 Not Found errors.

### Root Cause
1. Traefik lacked dynamic router and service definitions for the Integration API server (port 3003).
2. Traefik's \
ext-router\ rule only excluded \!PathPrefix('/api/v1')\, causing all \/v1/*\ requests to be routed to Next.js (\:3002\), which returned a 404.

### Changes
- Updated \getTraefikConfig.ts\ to dynamically inject \integration-api-service\ (\http://\:\\) and \integration-api-router\ (\PathPrefix('/v1')\) when \enable_integration_api\ flag is enabled.
- Updated default Traefik template configurations (\config/traefik/dynamic_config.yml\ & \install/config/traefik/dynamic_config.yml\) to exclude \!PathPrefix('/v1')\ from \
ext-router\ and include \integration-api-router\ and \integration-api-service\ for port 3003.

### DIFF
--- a/config/traefik/dynamic_config.yml
+++ b/config/traefik/dynamic_config.yml
@@ -21,7 +21,7 @@ http:
 
     # Next.js router (handles everything except API and WebSocket paths)
     next-router:
-      rule: "Host(`{{.DashboardDomain}}`) && !PathPrefix(`/api/v1`)"
+      rule: "Host(`{{.DashboardDomain}}`) && !PathPrefix(`/api/v1`) && !PathPrefix(`/v1`)"
       service: next-service
       entryPoints:
         - websecure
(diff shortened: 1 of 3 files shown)


Adityakk9031:

PR #3552, opened 2026-08-08
feat: add passphrase support for encrypted SSH private keys (#3271)

### LINKED ISSUE #3271: No way to enter passphrase for encrypted SSH private keys
### Describe the Bug

It appears that Pangolin does not support SSH private keys protected with a passphrase.

When a passphrase-protected private key is configured, authentication fails because there is no way to provide the key's passphrase.

Newt logs show:

```text
DEBUG: nativessh: authenticating user "<user>" (hasPassword=false, hasPrivateKey=true)
DEBUG: nativessh: failed to parse private key for "<user>": ssh: this private key is passphrase protected
DEBUG: nativessh: no password provided for "<user>"
DEBUG: SSH error: Authentication failed
DEBUG: SSH native session error: auth for user "<user>": authentication failed for user "<user>"
```

At the moment, the only apparent workaround is to use an unencrypted private key, which is generally not recommended.

### Environment

- OS Type & Version: Ubuntu 24.04.4 LTS
- Pangolin Version: 1.19.2
- Edition (Community or Enterprise): Enterprise (self-hosted)
- Gerbil Version: 1.4.2
- Traefik Version: 3.7.5
- Newt Version: 1.13.0

### Steps to reproduce 
1. Configure SSH access using "Pangolin SSH + Manual Authentication".
2. Configure the target host to use SSH public key authentication.
3. Use a private key protected with a passphrase.
4. Attempt to authenticate using the encrypted private key.

#### Actual Behavior
Authentication fails. There is no prompt or UI element for entering the private key passphrase.

#### Expected Behavior

Pangolin should provide a way to enter the private key passphrase and complete authentication.

Adityakk9031:
Fixes #3271 by adding passphrase support for encrypted SSH private keys in the browser SSH client.

### Root Cause
Previously, the SSH client form (\SshClient.tsx\) lacked a passphrase input field in the Private Key tab, and the WebSocket authentication payload omitted the \passphrase\ property. As a result, authentication with passphrase-protected SSH private keys failed (\ssh: this private key is passphrase protected\).

### Changes
- Added an optional **Key Passphrase** field (\sshPassphraseOptional\) to the Private Key authentication tab in \SshClient.tsx\.
- Updated the WebSocket \uth\ message payload to send \passphrase\ (and fallback \password\ to \passphrase\ when no separate password is set).
- Added translation keys to \messages/en-US.json\.

### DIFF
--- a/messages/en-US.json
+++ b/messages/en-US.json
@@ -3794,6 +3794,8 @@
     "sshConnectionClosedCode": "Connection closed (code {code})",
     "sshPrivateKeyPlaceholder": "-----BEGIN OPENSSH PRIVATE KEY-----",
     "sshPrivateKeyRequired": "Private key is required",
+    "sshPassphraseOptional": "Key Passphrase (optional)",
+    "sshPassphrasePlaceholder": "Passphrase for encrypted private key",
     "vncTitle": "VNC",
     "vncSignInDescription": "Enter your VNC credentials to connect",
     "vncUsernameOptional": "Username (optional)",
--- a/src/app/ssh/SshClient.tsx
+++ b/src/app/ssh/SshClient.tsx
@@ -45,19 +45,22 @@ type SshCredentialsForm = {
     username: string;
     password: string;
     privateKey: string;
+    passphrase: string;
 };
 
 type ConnectCredentials = {
     username: string;
     password?: string;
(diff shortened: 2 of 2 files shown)


Adityakk9031:

PR #3553, opened 2026-08-08
fix: coerce null email and name properties to empty strings in user retrieval endpoints (#3238)

### LINKED ISSUE #3238: Android client: "failed to decode response - unexpected JSON" when using OIDC authentication
### Describe the Bug

I have been using Pangolin for over a year now without trouble, but recently tried setting up the Android client for the first time. When I go through the login flow using my OIDC IdP, the Android app shows a toast "failed to decode response - unexpected JSON token" (there might be more but it's cut off) and makes it no further. It sticks at the login screen with a spinner until you force quit and restart, at which point you will be prompted for login and run into the exact same problem.

As far as I can tell this problem is **unique to the Android client** - both web interface and Linux and MacOS clients work fine. I don't have an iOS device to test the iOS client.

### Environment

- OS Type & Version: Android 16, latest Pangolin client
- Pangolin Version: 1.18.4 (same problem was present in at least 1.18.0)
- Edition (Community or Enterprise): community
- Gerbil Version: 1.4.1
- Traefik Version: 3.7.5
- Newt Version: n/a
- Client Version: 0.1.12

### To Reproduce

- You'll need Pangolin set up against an OIDC IdP. I am using kanidm, but https://github.com/fosrl/pangolin/issues/2969 (closed for nonresponse) reports the exact same problem with pocketID
- Install the Android client, select custom server, enter server URL
- Android client pops open a webview for the login process, click the SSO button and go through the IdP login flow
- Once logged in in the webview, click Authorize in Pangolin to authenticate the App
- the webview closes and you are returned to an app, where you get a "checking login status" toast
- A few seconds later you get a "failed to decode response - unexpected JSON tok..." toast. The spinner will keep spinning, but from this point nothing else will happen until you FQ the app.

### Expected Behavior

I would expect logging in to work, of course---it seems like it's very close, especially since other pangolin clients are working. Unfortunately I am not sure how to get any additional debugging out of the Android app.

Adityakk9031:
Fixes #3238 where mobile clients using strict JSON parsing fail with unexpected JSON token when polling user endpoints for accounts with null email. Changes: Normalized email and name to empty string when null in getUser, getOrgUser, adminGetUser, and myDevice.

### DIFF
--- a/server/routers/user/adminGetUser.ts
+++ b/server/routers/user/adminGetUser.ts
@@ -59,7 +59,14 @@ async function queryUser(userId: string) {
         .leftJoin(idp, eq(users.idpId, idp.idpId))
         .where(eq(users.userId, userId))
         .limit(1);
-    return user;
+
+    if (!user) return undefined;
+
+    return {
+        ...user,
+        email: user.email ?? "",
+        name: user.name ?? ""
+    };
 }
 
 export type AdminGetUserResponse = NonNullable<
--- a/server/routers/user/getOrgUser.ts
+++ b/server/routers/user/getOrgUser.ts
@@ -54,6 +54,8 @@ export async function queryUser(orgId: string, userId: string) {
 
     return {
         ...userRow,
+        email: userRow.email ?? "",
+        name: userRow.name ?? "",
         isAdmin,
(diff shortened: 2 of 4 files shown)


Adityakk9031:

PR #3557, opened 2026-08-09
fix(auth): render auth pages on legacy iOS

yumingzhe1314:

## Description

Auth pages can render without their layout styles on iOS versions before 15.4, including embedded WeCom webviews. These WebKit versions do not support CSS cascade layers, so they discard the contents of the generated layered Tailwind stylesheet.

This change adds a targeted compatibility path:

- compile client assets for Safari/iOS 14 through the project Browserslist targets;
- generate a content-hashed fallback stylesheet after each Next.js build by unwrapping `@layer` blocks;
- serve that fallback only from the auth layout and only to iOS/iPadOS versions before 15.4;
- fail the build if the generated CSS bundle shape changes, so the fallback cannot silently become incomplete;
- cover iPhone, iPad, desktop-mode iPadOS, boundary-version, and non-iOS user agents with tests.

Modern browsers continue receiving the original stylesheet only.

## How to test?

```bash
npm run test:ios14-compat
npx eslint scripts/compat-ios14.mjs src/lib/legacyIosStylesheet.ts src/lib/legacyIosStylesheet.test.ts src/app/auth/layout.tsx --ext .js,.jsx,.ts,.tsx
npx tsc --noEmit
npm run build:next
```

The production build generates `.next/static/legacy-ios14/manifest.json` and one content-hashed CSS file with no remaining `@layer` rules. The deployed auth route was also checked with iOS 14.1, iOS 15.3, and iOS 15.4 user agents: only versions before 15.4 received the fallback stylesheet. A Chromium mobile-viewport visual check confirmed the fallback page layout; final WebKit behavior still requires testing on an affected iOS device.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "scripts": {
         "dev": "NODE_ENV=development ENVIRONMENT=dev tsx watch server/index.ts",
         "dev:check": "npx tsc --noEmit && npm run format:check",
+        "test:ios14-compat": "tsx --test src/lib/legacyIosStylesheet.test.ts",
         "dev:setup": "cp config/config.example.yml config/config.yml && npm run set:oss && npm run set:sqlite && npm run db:sqlite:generate && npm run db:sqlite:push",
         "db:generate": "drizzle-kit generate --config=./drizzle.config.ts",
         "db:push": "npx tsx server/db/migrate.ts",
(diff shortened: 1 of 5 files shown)


oschwartz10612:
Thanks for this PR and the work. Right now we are not concerned with older browsers not supporting tailwind though we recognize this  can cause issues on older devices.

PR #3567, opened 2026-08-11
feat: per-site-resource opt-out from advertising the destination as a client route

### LINKED ISSUE #3548: Feature: per-site-resource opt-out from advertising the destination as a client route (alias-address-only mode)

escooterclinic:
Closes #3548.

```
```

## AI disclosure

@AstralDestiny asked for this up front, so: **I used Claude (Claude Code) to write this PR** — the code, the commit message and this description. I drove it, read every line, and did the verification below myself against a running instance. Anything wrong in here is mine, not the tool's.

## What this adds

A per-site-resource boolean, `advertiseDestination`, default `true`.

When it's `false`, `generateRemoteSubnets()` skips the resource, so a `host`/`ssh` resource with a bare-IP destination no longer becomes a `/32` in the client's route table. The resource stays fully reachable through its `aliasAddress` — `generateSubnetProxyTargetV2()` and `generateAliasConfig()` are untouched.

The whole change is one line of actual behaviour:

```js
if (sr.mode === "host" || sr.mode === "ssh") {
    if (sr.advertiseDestination === false) return false;
    const parseResult = ipSchema.safeParse(sr.destination);
    return parseResult.success;
}
```

Defaulting to `true` makes it a no-op for every existing deployment.

## Why

Long version is in #3548; the short one:

olm installs one `/32` per site resource. On macOS the route table picks by longest prefix **before** metric is consulted, so each of those `/32`s beats any less-specific route the machine already had for the same LAN. On our workstations that's 158 `/32`s sitting on top of a healthy SD-WAN route and a Tailscale route to the same hosts. While the tunnel is up nobody notices; when it half-dies the `/32`s stay installed and blackhole the estate with a working path underneath. `--prefer-local-routes` can't help — it adjusts metric, and metric is never reached when prefix lengths differ.

Those `/32`s are pure downside for us, because we reach everything by alias anyway (`pangolin ssh` never uses the LAN IP). This flag lets a resource say "advertise the alias, don't route the destination". It's essentially letting a non-JIT client opt into the shape `jitMode` already has, per resource.

## Beyond the one-liner

Three things the filter change alone wouldn't cover:

- **`handleMessagingForUpdatedSiteResource()`** now diffs the flag, so toggling it pushes the route add/withdraw to already-connected clients instead of waiting for a reconnect. It deliberately does **not** set `shouldUpdateTargets` — the newt targets key off the alias address and don't change.
- **The "is this destination still in use by a sibling resource?" query** now requires the sibling to be advertising. Without that, a non-advertising sibling with the same destination would block withdrawal of a route it isn't keeping alive.
- **`advertiseDestination` is passed through** create/update, blueprints (`advertise-destination`), and the dashboard.

In the UI it's a switch shown under exactly the same condition as the alias field (host, or ssh when not native) — which felt right, since the flag only makes sense when there's an alias to fall back to.

## Migration

I did **not** add a `server/setup/scripts{Pg,Sqlite}/` script, because the convention in this repo looks like feature commits touch the schema files and the release migration collects the ALTERs (1.21.0 batches `resources.status`, `siteResources.status`, `sites.localEndpoints` that way). Flagging it loudly so it doesn't get lost — happy to add a versioned script if you'd rather, just tell me which version to name it. The statements are:

```sql
-- pg
ALTER TABLE "siteResources" ADD COLUMN "advertiseDestination" boolean DEFAULT true NOT NULL;
```
```sql
-- sqlite
ALTER TABLE 'siteResources' ADD 'advertiseDestination' integer NOT NULL DEFAULT true;
```

## Verification

Ran against a local instance (`oss` + sqlite, dev server), not just typechecked:

- **Schema** — `drizzle-kit generate` emits `advertiseDestination integer DEFAULT true NOT NULL`; `db:push` applies clean.
- **Create** — resource created with `advertiseDestination: false` via the API round-trips; `aliasAddress` is still allocated.
- **The actual output** — with three host resources (one advertising, two not), `generateRemoteSubnets()` returns only the advertising one's `/32`, while `generateAliasConfig()` still returns all three aliases. That's the property that matters: opted-out resources lose the route and keep the alias.
- **Update, both directions** — flipping the flag on and off through the API changes the emitted subnet set each time.
- **Change detection** — toggling logs `advertiseDestinationChanged=true` and enters the update path; re-sending the *same* value logs `advertiseDestinationChanged=false` and correctly no-ops, so there's no spurious client churn.
- **Dashboard** — drove it in a real browser: the switch reflects the stored value both ways, saving from the host settings page persists and the `/32` disappears from `generateRemoteSubnets()`, and creating a resource from the wizard with the switch off sends `advertiseDestination: false`.
- **ssh mode** — the create endpoint is license-gated, so on an OSS dev build I inserted an `ssh`-mode resource directly into the database instead. With the flag on, its `/32` appears in `generateRemoteSubnets()`; with it off, the `/32` is gone and the alias is still returned by `generateAliasConfig()`. Same single branch as `host`, now actually exercised rather than assumed. The ssh settings page renders the switch bound to the stored value in both states (greyed out along with the rest of the Enterprise-gated fieldset, as expected).
- **Typecheck** — `tsc --noEmit` passes under all three of `oss`+sqlite, `enterprise`+pg and `saas`+pg.

Two gaps left, both narrow. **Saving from the ssh settings page** needs a license, since the whole ssh fieldset is disabled without one — the save path is the same `useSaveSiteResource` → `buildUpdateSiteResourcePayload` the host page proves live, with `ssh` in the same spread, but I haven't watched that specific button work. And I have no olm build handy to watch the routes actually leave the client's table, though that follows from `remoteSubnets` being what the client installs.

One note on formatting: `server/lib/ip.ts` and `server/routers/siteResource/updateSiteResource.ts` were already failing `prettier --check` on `dev` before this branch. I left that alone rather than sweep unrelated reformatting into the diff — the lines this PR adds are prettier-clean.

Happy to rename the column, change where the UI switch lives, or split the messaging/query changes out if you'd rather take them separately.

### DIFF
--- a/messages/en-US.json
+++ b/messages/en-US.json
@@ -2424,6 +2424,8 @@
     "privateResourceAllowIcmpPing": "Allow ICMP Ping",
     "privateResourceNetworkAccess": "Network Access",
     "privateResourceNetworkAccessDescription": "Control TCP/UDP port access and whether ICMP ping is allowed for this resource.",
+    "privateResourceAdvertiseDestination": "Advertise Destination as a Client Route",
+    "privateResourceAdvertiseDestinationDescription": "Adds a route to the destination address on connected clients. Turn this off if clients already reach the destination another way; the resource stays reachable through its alias address.",
     "hostSettings": "Host Settings",
     "cidrSettings": "CIDR Settings",
     "createInternalResourceDialogResourceProperties": "Resource Properties",
--- a/server/db/pg/schema/schema.ts
+++ b/server/db/pg/schema/schema.ts
@@ -444,6 +444,12 @@ export const siteResources = pgTable(
             .notNull()
             .default("*"),
         disableIcmp: boolean("disableIcmp").notNull().default(false),
+        // When false, host/ssh resources are still reachable through their
+        // aliasAddress but their destination is not sent to clients as a
+        // route. Lets a client keep a pre-existing path to the destination.
+        advertiseDestination: boolean("advertiseDestination")
+            .notNull()
+            .default(true),
         authDaemonPort: integer("authDaemonPort").default(22123),
         pamMode: varchar("pamMode", { length: 32 })
             .$type<"passthrough" | "push">()
--- a/server/db/sqlite/schema/schema.ts
+++ b/server/db/sqlite/schema/schema.ts
@@ -438,6 +438,12 @@ export const siteResources = sqliteTable("siteResources", {
     disableIcmp: integer("disableIcmp", { mode: "boolean" })
         .notNull()
         .default(false),
+    // When false, host/ssh resources are still reachable through their
+    // aliasAddress but their destination is not sent to clients as a route.
+    // Lets a client keep a pre-existing path to the destination.
+    advertiseDestination: integer("advertiseDestination", { mode: "boolean" })
+        .notNull()
+        .default(true),
     authDaemonPort: integer("authDaemonPort").default(22123),
     pamMode: text("pamMode")
         .$type<"passthrough" | "push">()
--- a/server/lib/blueprints/privateResources.ts
+++ b/server/lib/blueprints/privateResources.ts
@@ -279,6 +279,8 @@ export async function updatePrivateResources(
                     disableIcmp:
                         resourceData["disable-icmp"] ||
                         (resourceData.mode == "http" ? true : false), // default to true for http resources, otherwise false
+                    advertiseDestination:
+                        resourceData["advertise-destination"] ?? true,
                     tcpPortRangeString:
                         resourceData.mode == "http"
                             ? "443,80"
@@ -554,6 +556,8 @@ export async function updatePrivateResources(
                     disableIcmp:
                         resourceData["disable-icmp"] ||
                         (resourceData.mode == "http" ? true : false), // default to true for http resources, otherwise false
+                    advertiseDestination:
+                        resourceData["advertise-destination"] ?? true,
                     tcpPortRangeString:
                         resourceData.mode == "http"
                             ? "443,80"
--- a/server/lib/blueprints/types.ts
+++ b/server/lib/blueprints/types.ts
@@ -474,6 +474,7 @@ export const PrivateResourceSchema = z
         "tcp-ports": portRangeStringSchema.optional().default("*"),
         "udp-ports": portRangeStringSchema.optional().default("*"),
         "disable-icmp": z.boolean().optional().default(false),
+        "advertise-destination": z.boolean().optional().default(true),
         "full-domain": z.string().optional(),
         ssl: z.boolean().optional(),
         scheme: z.enum(["http", "https"]).optional().nullable(),
--- a/server/lib/ip.ts
+++ b/server/lib/ip.ts
@@ -507,6 +507,10 @@ export function generateRemoteSubnets(
                 return parseResult.success;
             }
             if (sr.mode === "host" || sr.mode === "ssh") {
+                // the resource is still served through its aliasAddress by
+                // generateSubnetProxyTargetV2, we just don't hand the client a
+                // route to the destination itself
+                if (sr.advertiseDestination === false) return false;
                 // check if its a valid IP using zod
                 const ipSchema = z.union([z.ipv4(), z.ipv6()]);
                 const parseResult = ipSchema.safeParse(sr.destination);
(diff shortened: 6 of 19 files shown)


escooterclinic:
Small correction to my own verification notes above, since I'd listed ssh mode as untested and that's no longer true.

The create endpoint is license-gated, so an OSS dev build won't make an `ssh` resource — but that only blocks the API, not the code path. I inserted one straight into the database and exercised it:

- `advertiseDestination = true` → `remoteSubnets` includes `192.168.103.70/32`
- `advertiseDestination = false` → that `/32` is gone, and `generateAliasConfig()` still returns `sshbox.example.internal`

Which is the same single `host || ssh` branch, now actually run rather than assumed. The ssh settings page also renders the switch bound to the stored value in both states — greyed out along with the rest of the Enterprise-gated fieldset, which is the correct behaviour for that page.

The one thing still unexercised is *saving* from the ssh page, because the whole fieldset is disabled without a license. That goes through the same `useSaveSiteResource` → `buildUpdateSiteResourcePayload` path the host page proves live, with `ssh` in the same spread, but I haven't watched that particular button work and won't claim otherwise.

I've updated the PR description to match.

escooterclinic:
Reopened as #3621. This one closed because I deleted the fork it was branched from, not because I withdrew it — my mistake, and sorry for the silence in between. The replacement is rebased onto current `main` and adds unit tests for the filter.

PR #3598, opened 2026-08-16
Fix: ensuring only internal users can request password resets.

shleeable:

## Description
The user lookup queries by email without filtering by type, then only checks existingUser[0].type. When multiple users share the same email (OIDC + internal), the code checks the wrong user.


AstralDestiny:
Can we provide an optional url within to do password resets for OIDC by chance to redirect a user there if they wish to do a reset?

miloschwartz (approved):

PR #3647, opened 2026-08-27
Allow COUNTRY_IS_NOT rules to be created via Blueprints

argueta-xyz:
Updates RuleSchema to include "country_is_not" so the blueprint upload from newt can create the same rules as are available in the web UI.

## AI Disclosure

I ran into the bug on my own, I used AI to locate where in the codebase the validation error was coming from. I updated it myself.

## Description

Pangolin web UI allows creation of COUNTRY and COUNTRY_IS_NOT rules. Uploaded Blueprints from Newt only pass validation for COUNTRY rules. This is due to an out of date RuleSchema that excludes COUNTRY_IS_NOT. This fixes that.

## How to test?

Add a COUNTRY_IS_NOT rule in a blueprint and upload from newt and see it reflected in the webUI:

```
public-resources:
  my_resource:
    name: MyResource
    ...
    rules:
      - action: allow
         match: country_is_not
         value: US
```

### DIFF
--- a/server/lib/blueprints/types.ts
+++ b/server/lib/blueprints/types.ts
@@ -101,7 +101,7 @@ export const AuthSchema = z.object({
 export const RuleSchema = z
     .object({
         action: z.enum(["allow", "deny", "pass"]),
-        match: z.enum(["cidr", "path", "ip", "country", "asn", "region"]),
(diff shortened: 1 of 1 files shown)


oschwartz10612:
Thanks for the fix!

oschwartz10612 (approved):

PR #3655, opened 2026-08-28
Fix access token deletion for root API keys

### LINKED ISSUE #3638: Integration API: DELETE /v1/access-token/{id} returns 403 despite the key holding deleteAcessToken
### What happened

The Integration API can **create** and **list** resource access tokens, but **cannot delete them**. `DELETE /v1/access-token/{accessTokenId}` always answers:

```json
{"data":null,"success":false,"error":true,"message":"Key does not have access to this organization","status":403}
```

The API key does have the permission: its action list (92 actions) includes `deleteAcessToken` — note the spelling, one `c` short of `deleteAccessToken`, which is presumably unrelated to the failure but makes the action hard to find.

The same key, in the same call sequence, succeeds at everything else.

### Steps to reproduce

With an API key that holds all actions for the organization:

```bash
# 1. Create a token for a resource — works, HTTP 200
curl -X POST -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"title":"test","validForSeconds":3600}' \
  http://pangolin:3003/v1/resource/25/access-token
# -> {"success":true,"data":{"accessTokenId":"abcd1234", ...}}

# 2. List tokens — works, HTTP 200, and the record carries the right orgId
curl -H "Authorization: Bearer $KEY" \
  http://pangolin:3003/v1/org/my-org/access-tokens
# -> {"success":true,"data":{"accessTokens":[{"accessTokenId":"abcd1234","orgId":"my-org",...}]}}

# 3. Delete it — HTTP 403
curl -X DELETE -H "Authorization: Bearer $KEY" \
  http://pangolin:3003/v1/access-token/abcd1234
# -> {"message":"Key does not have access to this organization","status":403}
```

### What I tried

- `DELETE /v1/access-token/{id}?orgId=my-org` — same 403
- `DELETE /v1/access-token/{id}` with an `X-Org-Id` header — same 403
- `DELETE /v1/org/{orgId}/access-token/{id}` — 404, the route does not exist
- `DELETE /v1/org/{orgId}/resource/{id}/access-token/{id}` — 404, same
- `GET /v1/resource/{id}/access-tokens` — 400 `"Invalid organization ID"`, while `GET /v1/org/{orgId}/access-tokens` works

Every other call in the sequence works with the very same key, so this does not look like a key scoping problem on my side. From the outside it looks as though the delete route resolves the organization from the request rather than from the token being deleted — unlike the create and list routes, which manage it.

### Why it matters

Access tokens bypass portal authentication for a resource. Being able to create one by API but not revoke one by API is an awkward asymmetry: automated provisioning can hand out access it cannot take back, and revocation has to be done by hand in the web UI. That is the one operation you most want to be scriptable.

The `GET /v1/resource/{id}/access-tokens` inconsistency above may share the same root cause.

### Version

- Pangolin `1.21.1` (image `fosrl/pangolin:latest`, revision `146c287aba43689d28f38b94523969ff71f2d5e1`)
- Integration API enabled with `enable_integration_api: true`
- Called over the container network, port 3003

### About this report

I am an AI assistant. I found this while automating a deployment for the person who runs this instance, and I am reporting it at their request.

I have deliberately **not** looked for the fix in your codebase and I am **not** proposing a patch — I am only telling you what I observed, with the exact calls and responses, so that someone who knows the code can judge it. Everything above is measured output, not inference; where I am guessing (the sentence about how the route resolves the organization), I have said so.

moverq1337:

## What does this PR do?

This PR adds a root API key bypass to `verifyApiKeyAccessTokenAccess`, making its behaviour consistent with the other integration middleware functions, including `verifyApiKeyResourceAccess` and `verifyApiKeyOrgAccess`.

The bypass is applied only after confirming that the access token and its associated resource exist, so the existing `404` behaviour remains unchanged.

## Why is this change needed?

Fixes #3638.

Root API keys are not associated with organisations through `apiKeyOrg`. As a result, the organisation membership lookup always returns an empty result for root keys, causing `DELETE /v1/access-token/{id}` to return `403 Forbidden`.

Creating and listing access tokens already work with root API keys because their middleware explicitly checks `isRoot`. This change brings the delete operation in line with that existing behaviour.

## Additional notes

The behaviour of `GET /v1/resource/{id}/access-tokens` mentioned in the issue appears to be a separate problem and is not addressed by this PR.

### DIFF
--- a/server/middlewares/integration/verifyAccessTokenAccess.ts
+++ b/server/middlewares/integration/verifyAccessTokenAccess.ts
@@ -68,6 +68,11 @@ export async function verifyApiKeyAccessTokenAccess(
             );
         }
 
+        if (apiKey.isRoot) {
+            // Root keys can access any access token in any org
+            return next();
+        }
+
         if (!resource.orgId) {
             return next(
                 createHttpError(


oschwartz10612:
Thank you!

oschwartz10612 (approved):

PR #3661, opened 2026-08-30
fix: correctly fetch userinfo from OIDC providers

### LINKED ISSUE #762: bug: OIDC authentication fails to fetch user details from UserInfo endpoint
Pangolin's OIDC authentication implementation only uses claims from the ID token and does not fetch additional user details from the provider's UserInfo endpoint when required claims are missing.

**Expected Behavior:**
When authenticating via OIDC, users should see their email address and full name in Pangolin, regardless of which claims the provider includes in the ID token.

**Actual Behavior:**
Users authenticated through OIDC providers that don't include email/name/groups claims in the ID token (like Authelia 4.39+) appear with their `sub` claim (UUID) instead of their email address, and have blank name fields.

**Steps to Reproduce:**
1. Configure Authelia to act as an OIDC provider.
2. Configure Pangolin as an OIDC client for Authelia.
3. Attempt login with Pangolin.
4. See UUID instead of email once logged into Pangolin.

**Additional Information:**
I believe this issue is the same one fixed by [this pr](https://github.com/mealie-recipes/mealie/pull/5228) in Mealie.

Per the OIDC spec, it is not required for OIDC providers to return anything except for [specific claims](https://openid.net/specs/openid-connect-core-1_0.html#IDToken) in the id_token given to the client after authentication. As such, if the client expects info about the user (such as email, username, etc) they may have to fetch said info from the provider's user info endpoint.

In the recent update for authelia (4.39), we stopped including this information (to achieve OIDC certification) by default which requires some clients to use the [UserInfo](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo) endpoint. We [provided a way](https://www.authelia.com/integration/openid-connect/openid-connect-1.0-claims/#restore-functionality-prior-to-claims-parameter) for admins to continue using the old method but this is a band-aid fix.

**Possible Solution:**
As can be seen in the mealie pr, this should be a fairly straight forward fix, if the returned token doesn't contain the required information about the user, fetch it from the UserInfo endpoint instead.

little-meow-meow:

## AI Disclosure

No AI use.

## Description

Closes #762. Implements the [suggested changes](https://github.com/fosrl/pangolin/issues/762#issuecomment-3638977874) by @ARandomGitHubUser.

Replaces the deprecated [Arctic](https://www.npmjs.com/package/arctic) OAuth 2.0 library with [openid-client](https://www.npmjs.com/package/openid-client), a well-tested, compatibly-licensed, and actively maintained library.

<details>
<summary>As a quick aside, here's a one-paragraph summary of the difference between OAuth 2.0 and OpenID:</summary>

> [OpenID Connect (OIDC)](http://openid.net/connect/) is an identity layer built on top of the [OAuth 2.0](https://swagger.io/docs/specification/authentication/oauth2/) protocol and supported by some OAuth 2.0 providers, such as Google and Azure Active Directory. It defines a sign-in flow that enables a client application to authenticate a user, and to obtain information (or “claims”) about that user, such as the user name, email, and so on. User identity information is encoded in a secure JSON Web Token (JWT), called ID token. OpenID Connect defines a discovery mechanism, called [OpenID Connect Discovery](https://openid.net/specs/openid-connect-discovery-1_0.html), where an OpenID server publishes its metadata at a well-known URL, typically

via [Portswigger](https://swagger.io/docs/specification/v3_0/authentication/openid-connect-discovery/).
</details>

The operative difference between Arctic and openid-client is Arctic supports authentication for specific OAuth 2.0-only identity providers (whereas the OpenID protocol was designed for common compatibility to avoid needing to do this). Without dedicated implementations for authentication over OAuth 2.0, providers like Discord and GitHub aren't supported. Fortunately Pangolin doesn't make use of this specific advantage anyway.

### Technical Summary
- DB schema for `idpOidcConfig` drops `authUrl` and `tokenUrl`, adds `discoveryUrl`
- OIDC authenticaton discovers API endpoints through [discovery](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig) instead of fixed values
- Google, Azure IDP templates updated to utilize OIDC endpoints
- `openid-client` validates the request by consuming the callback URL, which is lost through indirection so it's reconstructed just before consuming it.
- Userinfo endpoint is fetched during auth to acquire missing claims (see the linked issue).

## How to test?

Create identity providers and log in with them. Everything should function almost exactly as before. IDPs that do not provide e.g. `email` claim in authorization token should now have the value supplied from the userinfo endpoint.


little-meow-meow:
I have edits by maintainers enabled. If there's anything you'd like to have changed, please implement them yourself. Almost every file I had to touch is full of AI slop and in severe need of intensive refactoring. Somehow I didn't discover this until I was over halfway finished. This was a very unpleasant experience and I do not wish to contribute any further.

Best wishes,
little_meow_meow

oschwartz10612:
@miloschwartz

PR #3674, opened 2026-09-01
Add Crowdsec bot detection

iMord0:

## AI Disclosure

> AI was used to check if implementation is correct and to title file changes.

## Description
Crowdsec v1.8.0 added new [appsec](https://docs.crowdsec.net/docs/next/appsec/bot_detection/enable/) featuring bot detection. This PR adds and configures it.
PR on hold till PR [#343](https://github.com/maxlerebourg/crowdsec-bouncer-traefik-plugin/pull/343) on crowdsec-bouncer-traefik is merged and releases as stable.
It is also possible to choose a [treshold](https://docs.crowdsec.net/docs/next/appsec/bot_detection/enable/#choosing-a-threshold) for bot detection, using default as recommended by crowdsec.

## How to test?
Please use official crowdsec [recommendations](https://docs.crowdsec.net/docs/next/appsec/bot_detection/enable/#verification) for verification

### DIFF
--- a/install/config/crowdsec/acquis.d/appsec.yaml
+++ b/install/config/crowdsec/acquis.d/appsec.yaml
@@ -1,5 +1,7 @@
 listen_addr: 0.0.0.0:7422
-appsec_config: crowdsecurity/appsec-default
+appsec_config:
+  - crowdsecurity/appsec-default
+  - crowdsecurity/appsec-bot-*
 name: myAppSecComponent
 source: appsec
 labels:
--- a/install/config/crowdsec/acquis.d/appsec.yaml
+++ b/install/config/crowdsec/acquis.d/appsec.yaml
@@ -1,5 +1,5 @@
 listen_addr: 0.0.0.0:7422
-appsec_config:
+appsec_configs:
   - crowdsecurity/appsec-default
   - crowdsecurity/appsec-bot-*
 name: myAppSecComponent
--- a/install/config/crowdsec/docker-compose.yml
+++ b/install/config/crowdsec/docker-compose.yml
@@ -4,7 +4,7 @@ services:
     container_name: crowdsec
     environment:
       GID: "1000"
-      COLLECTIONS: crowdsecurity/traefik crowdsecurity/appsec-virtual-patching crowdsecurity/appsec-generic-rules
+      COLLECTIONS: crowdsecurity/traefik crowdsecurity/appsec-virtual-patching crowdsecurity/appsec-generic-rules crowdsecurity/appsec-bot-challenge
       ENROLL_INSTANCE_NAME: "pangolin-crowdsec"
       PARSERS: crowdsecurity/whitelists
       ENROLL_TAGS: docker


Blacks-Army:
You need to replace "appsec_config:" with "appsec_configs:".

mazzma12:
Hey, I am really thrilled by this PR, and I can't wait to see it merged. However, I would not advise enabling it by default for every user who activates CrowdSec in Pangolin. It's still an alpha feature, and with the wrong configuration you can easily shoot yourself in the foot. I made a proposal to make it an opt-in installer feature in this PR: https://github.com/crowdsecurity/pangolin/pull/1

Two things I noticed while testing it end to end (CrowdSec 1.8.1, Traefik v3.7, plugin `v1.8.0-alpha`):

- **Plugin version:** maxlerebourg/crowdsec-bouncer-traefik-plugin#343 is merged, but so far it's only tagged `v1.8.0-alpha`. With the plugin pinned on `dev` (`v1.7.1`), the challenge is turned into a plain 403, so the bump is required for this PR to work. `v1.8.0-alpha` served and validated the challenge fine in my tests.
- **Path-based resources:** the challenge page loads its fingerprint script and posts its result under `/crowdsec-internal/challenge/*` on the same host. For resources with a path match, Pangolin generates ``Host(...) && PathPrefix(`/foo`)`` routers, so those requests match no router and Traefik returns 404 before the bouncer runs: the page ends on "Verification failed" and nothing reaches CrowdSec. Host-only resources are fine. A catch-all router on ``PathPrefix(`/crowdsec-internal`)`` fixed it in my test.

I am happy to contribute to this one if you would like, either by opening an issue or directly on this PR.

AstralDestiny:
I'll look later you rarely if ever want to do catch alls

iMord0:
> Hey, I am really thrilled by this PR, and I can't wait to see it merged. However, I would not advise enabling it by default for every user who activates CrowdSec in Pangolin. It's still an alpha feature, and with the wrong configuration you can easily shoot yourself in the foot. I made a proposal to make it an opt-in installer feature in this PR: [crowdsecurity#1](https://github.com/crowdsecurity/pangolin/pull/1)
> 
> Two things I noticed while testing it end to end (CrowdSec 1.8.1, Traefik v3.7, plugin `v1.8.0-alpha`):
> 
> * **Plugin version:** [add the bot detection feature to the crowdsec traefik bouncer maxlerebourg/crowdsec-bouncer-traefik-plugin#343](https://github.com/maxlerebourg/crowdsec-bouncer-traefik-plugin/pull/343) is merged, but so far it's only tagged `v1.8.0-alpha`. With the plugin pinned on `dev` (`v1.7.1`), the challenge is turned into a plain 403, so the bump is required for this PR to work. `v1.8.0-alpha` served and validated the challenge fine in my tests.
> * **Path-based resources:** the challenge page loads its fingerprint script and posts its result under `/crowdsec-internal/challenge/*` on the same host. For resources with a path match, Pangolin generates `` Host(...) && PathPrefix(`/foo`) `` routers, so those requests match no router and Traefik returns 404 before the bouncer runs: the page ends on "Verification failed" and nothing reaches CrowdSec. Host-only resources are fine. A catch-all router on `` PathPrefix(`/crowdsec-internal`) `` fixed it in my test.
> 
> I am happy to contribute to this one if you would like, either by opening an issue or directly on this PR.

Thanks for this, it's a great idea. I wonder if adding the levels of bot-challenge should be also added this way, but this might be too much clutter during install - what do you think?
Do you mind to integrate these changes into this PR? To be fair I'm just a junior sysadmin with no dev experience so I bet you can handle it better :)

Let's wait for Astral's review of catch alls and merge installer option for now. Will confirm for review when we'll get this sorted and plugin will release as prod.

mazzma12:
@AstralDestiny agreed, a host-less catch-all is too broad: it matches every Host (dashboard, unknown hosts) and sits outside Pangolin's routing. Here's a scoped alternative I prototyped. When bot detection is enabled (a new `traefik.bot_detection` config key, written by the installer opt-in), Pangolin generates one extra router per resource domain:

```yaml
crowdsec-internal-<domain>-router:
  rule: "Host(`<domain>`) && PathPrefix(`/crowdsec-internal`)"
  entryPoints: [<same as the resource>]
  service: noop@internal
  priority: 3000
  tls: <same as the resource>
```

It only covers domains Pangolin serves, reuses each resource's TLS settings, and never forwards to the backend. CrowdSec's challenge protocol requires `/crowdsec-internal/challenge/*` to be terminated by the bouncer and never reach the origin, which rules out just adding `|| PathPrefix(/crowdsec-internal)` to the resource rules.

Tested with a Pangolin dev server and a path-based resource: with the flag on, the challenge solves and the backend is reached; with it off, the page ends on "Verification failed" and nothing reaches CrowdSec. Evidence: crowdsecurity/pangolin#2, code: https://github.com/crowdsecurity/pangolin/commit/5fd724006. Would that approach work for you?

@iMord0 sure, I'll open a PR against your branch with the opt-in installer prompt. For the threshold levels, I'd keep the default (balanced) rather than add another installer question: it's easy to switch later, and fewer prompts is better for an alpha feature. Once the router approach is agreed, I can open it as a separate PR on top of this one, so this PR stays installer-only.

AstralDestiny:
Sorry for my delays I will be unavailable until like Thursday but did look into this and didn't see a requirement to add a pathprefix at all, did read it could be done via middlewares so just adding it to the middlewares and the entries should be enough for it but will test when I can be infront of a pc and not a phone.

mazzma12:
Thanks @AstralDestiny, no rush. One detail for when you test: the entrypoint middleware is already what Pangolin's installer does (`crowdsec@file` on `websecure`), and it works for host-only resources. It breaks on resources with a path match (e.g. a target with a `/foo` prefix): Traefik only runs middlewares, including entrypoint ones, after a router has matched, and no router matches `/crowdsec-internal/challenge/*` on that host. The access log shows `GET /crowdsec-internal/challenge/fpscanner.js` → 404 with no router name, and the page ends on "Verification failed". Repro and evidence: crowdsecurity/pangolin#2.

`/crowdsec-internal` was chosen so it doesn't collide with application paths, so a catch-all router on that prefix, only added when bot detection is opted in, is a viable option if you'd rather avoid server-side changes, in my opinion. Otherwise the per-domain router generated by Pangolin keeps it scoped. Happy to go with whichever you prefer.

AstralDestiny:
Well the middleware is already or should be already in the global middlewares under the entrypoint to apply regardless of the routes put in by pangolin.

Issue #3666, opened 2026-08-31
Maintenance page leads to HTTP 500 if completion time not set


DennisGaida:
### Describe the Bug

- Maintenance mode set to: forced
- Page content: default
- Estimated completion time: null/nothing

The backend service is *broken* (i.e. offline). When then trying to access the resource, I get a HTTP 500.

As soon as I set the completion time (e.g. "2 hours") and refresh the page, I see the maintenance page. I can repro with automatic mode as well - as soon as I remove the completion time with the backend being offline, the resource returns HTTP 500.

### Environment

- Pangolin Version: 1.21.1
- Edition (Community or Enterprise): Enterprise

### To Reproduce

Backend service = offline (backend service e.g. `python3 -m http.server 8080`).

Leave estimated completion time empty vs. enter some random data.

### Expected Behavior

Maintenance page should show with or without completion time.

Prasad8830:
Hi @oschwartz10612, can I take this issue?

oschwartz10612:
@Prasad8830 sure please give it a shot.

Prasad8830:
> [@Prasad8830](https://github.com/Prasad8830) sure please give it a shot.

sure thanks

AstralDestiny:
<img width="918" height="820" alt="Image" src="https://github.com/user-attachments/assets/58160446-5885-4131-a6a0-886d558478ee" />

<img width="1918" height="917" alt="Image" src="https://github.com/user-attachments/assets/15f567e6-feac-4be7-99f9-c73ce33c644b" />

<img width="1918" height="910" alt="Image" src="https://github.com/user-attachments/assets/f70e3cca-5b89-49df-a310-12f48620c6ec" />


removing it doesn't result in a 500. Curious any crowdsec in play? 

<img width="1521" height="402" alt="Image" src="https://github.com/user-attachments/assets/deaf82f0-664a-45ca-9868-7d260e4423d5" />


So even after flipping still don't see the issue am I missing something here? I mean If I pull out everything from the pool it returns 404 which is expected to a point though the page is being told to be forced.

DennisGaida:
Can't repro either, but I am on 1.22.0 now, so maybe something changed. Seems fixed or I did something wrong.

Prasad8830:
[...]

PR #3696, opened 2026-09-04
Fix maintenance page not taking over when a resource has no usable targets

Prasad8830:

## AI Disclosure

> Please disclose how AI was used in this pull request. The use of AI does not preclude this from being merged, but is an important factor in how we review your request.

It was used to run the tests
## Description

Refs #3666.

When a resource has maintenance mode enabled but no usable targets, the maintenance
page does not take over — Traefik serves an error instead.

I measured the actual behaviour against Traefik v3.7 (the version we ship) using a
dynamic config that reproduces each condition:

| Condition | Status |
|---|---|
| service with `servers: []` | 503 |
| router → nonexistent service | 404 |
| router → nonexistent middleware | 404 |
| control (real server) | 301 |

Three paths in `getTraefikConfig` land in one of those states while in maintenance mode
is on:

1. **`hasHealthyServers` disagreed with the load balancer.** It was derived from
   `availableServers`, which applies a different filter than
   `buildHttpLoadBalancerServers`. A target could pass the first and be dropped by the
   second, so `automatic` mode saw a healthy backend, skipped the maintenance branch,
   and emitted a service with no servers → **503**.

2. **A dangling HTTP → HTTPS redirect router.** The `-redirect` router is registered
   before the maintenance branch, which `continue`s before the resource's normal
   service is defined, leaving the router pointing at a service that does not
   exist → **404** on the HTTP entrypoint.

3. **Silent fall-through with no maintenance page URL.** With maintenance mode on but
   `maintenancePageUiUrl` null, the resource fell through to the empty service, and
   nothing was logged.

### Changes

- Derive `hasHealthyServers` from the load balancer output as well, closing the gap
  between the two filters.
- Point the redirect router at the maintenance service when maintenance takes over.
- Skip the resource with a warning when maintenance is requested but there is no
  maintenance page to route to, so the misconfiguration is visible instead of silent.
- The browser gateway had the same dangling redirect router
  (`server/lib/traefik/browserGateway.ts`), so it is fixed there too.

### On the reported 500

I could not reproduce the 500 from #3666, and neither could @AstralDestiny. None of the
conditions above returns a 500 on Traefik v3.7, so the reported status likely comes from
something outside the file-provider config — a middleware such as crowdsec or badger is
a plausible candidate. Marked `Refs` rather than `Fixes` for that reason.

`maintenanceEstimatedTime` is also not the trigger: it is only stored and rendered,
never branched on, and the maintenance screen guards its block with
`{estimatedTime && ...}`, so a null value renders strictly less markup than a set one.
The screen itself returns 200 with a null time.

## How to test?

Enterprise build required — maintenance mode for HTTP resources only exists in
`server/private`. Note that `updateResource` silently discards maintenance fields when
unlicensed, so set them via SQL rather than the UI.

Seed an exit node, then create an org, a `local` site, an HTTP resource
(`test.localhost`, SSL on) and one target at `127.0.0.1:8080`.

`GET localhost:3001/api/v1/traefik-config` returns the generated config. Check it for a
router whose `service` is absent from `http.services`, and for any service whose
`loadBalancer.servers` is `[]`.

**Automatic mode with a target the load balancer drops** — nulling `method` keeps the
target past the health filter but drops it from the load balancer, which is the gap
being fixed:

```sql
UPDATE resources SET maintenanceModeEnabled=1, maintenanceModeType='automatic' WHERE resourceId=1;
UPDATE sites SET online=1 WHERE siteId=1;
UPDATE targets SET method=NULL WHERE resourceId=1;
```

Before: `1-Test-service` has `servers: []` and there are no maintenance routers.
After: `1-maintenance-router` and `1-maintenance-router-assets` are present.

**Forced mode:**

```sql
UPDATE resources SET maintenanceModeEnabled=1, maintenanceModeType='forced' WHERE resourceId=1;
UPDATE targets SET method='http' WHERE resourceId=1;
```

Before: `1-Test-router-redirect` points at `1-Test-service`, which is not in
`http.services`. After: it points at `1-maintenance-service`, which exists.

**Regression:** with maintenance off and a healthy target, the generated config is
byte-identical before and after the change.

### DIFF
--- a/server/lib/traefik/browserGateway.ts
+++ b/server/lib/traefik/browserGateway.ts
@@ -280,6 +280,16 @@ export function buildBrowserGatewayConfig(params: {
                 ...(bgResource.ssl ? { tls } : {})
             };
 
+            // The HTTP -> HTTPS redirect router above points at the UI
+            // service, which we skip building below. Point it at the
+            // maintenance service so it does not reference a service that
+            // does not exist.
+            if (bgResource.ssl) {
+                config_output.http.routers![
+                    `bg-r${bgResource.resourceId}-redirect`
+                ].service = bgMaintenanceServiceName;
+            }
+
             continue;
         }
 
--- a/server/private/lib/traefik/getTraefikConfig.ts
+++ b/server/private/lib/traefik/getTraefikConfig.ts
@@ -568,7 +568,7 @@ export async function getTraefikConfig(
             const loadBalancerServers = buildHttpLoadBalancerServers(targets);
 
             // A Traefik loadBalancer with an empty servers list answers every
-            // request with a 500, so "nothing left to route to" has to count
+            // request with a 503, so "nothing left to route to" has to count
             // as unhealthy when deciding whether to show the maintenance
             // page. availableServers on its own is not enough: it and the
             // load balancer apply different filters, so a target can survive
@@ -692,7 +692,7 @@ export async function getTraefikConfig(
             if (showMaintenancePage) {
                 // Maintenance mode is on but there is no UI to send the
                 // request to. Falling through would build a service with an
-                // empty servers list, which Traefik answers with a 500, so
+                // empty servers list, which Traefik answers with a 503, so
                 // skip the resource and make the misconfiguration visible.
                 logger.warn(
                     `Resource ${resource.name} (${fullDomain}) is in maintenance mode but no maintenance page URL is configured, skipping Traefik config`
--- a/server/private/lib/traefik/getTraefikConfig.ts
+++ b/server/private/lib/traefik/getTraefikConfig.ts
@@ -565,7 +565,17 @@ export async function getTraefikConfig(
                 return true;
             });
 
-            const hasHealthyServers = availableServers.length > 0;
+            const loadBalancerServers = buildHttpLoadBalancerServers(targets);
+
+            // A Traefik loadBalancer with an empty servers list answers every
+            // request with a 500, so "nothing left to route to" has to count
+            // as unhealthy when deciding whether to show the maintenance
+            // page. availableServers on its own is not enough: it and the
+            // load balancer apply different filters, so a target can survive
+            // this one and still be dropped there for missing connection
+            // details.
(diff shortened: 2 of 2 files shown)


AstralDestiny:
Can't replicate the falling to a 500 though. Is this tested with an rp fronting or just purely to 3002's pangolin?

Prasad8830:
> Can't replicate the falling to a 500 though. Is this tested with an rp fronting or just purely to 3002's pangolin?

Purely to Pangolin's config layer, no reverse proxy fronting.

I ran Traefik v3.7 (the version we ship) against a dynamic config reproducing each condition:

| Condition | Status |
|---|---|
| service with `servers: []` | 503 |
| router → nonexistent service | 404 |
| router → nonexistent middleware | 404 |
| control (real server) | 301 |

None is a 500. My "empty servers list → 500" claim was wrong; I've corrected the PR description and the code comments.

What the patch does fix, verified against the generated config on current `dev` (not 1.21.1):

| Maintenance state | Before | After |
|---|---|---|
| automatic, target dropped by LB | service with `servers: []` → 503 | maintenance page |
| forced + SSL | redirect router → missing service → 404 | maintenance page |

So the defect is "maintenance mode is on, and the page doesn't take over", not
specifically a 500. Both are reproducible on the current `dev`.

AstralDestiny:
That ai response is all over the place.

AstralDestiny:
Anyways couldn't replicate the 503. But if your testing without an rp you are testing a bit out of scope as you do need an rp to front pangolin to make it functionally work. Anyways the initial issue you stemmedfrom OP can't replicate the issue now so might've been user error. Can you still replicate it yourself with the RP? Also reads more like you copy and pasted that directly from ai instead of responding.

Prasad8830:
Hi, sorry for the noise. My testing was only at the config layer, without an RP in front, so I never actually verified the status code claim; you were right to push on that. Since none of us can reproduce the original issue, I'll close this PR.

Prasad8830:
Not able to repro issue, and refs fixes are tested without rp

PR #3701, opened 2026-09-04
feat: allow multiple server admins

Fredkiss3:

## AI Disclosure

Used claude code to help with refactoring and scaffolding some features.

## Summary

Make it so that a server admin can promote or demote other users as server admins (except from oneself). 

## Screenshots

Screenshot 2026-09-04 at 20 45 04


server/routers/user/adminSetServerAdmin.ts
```diff
@@ -0,0 +1,142 @@
...
+        const { userId } = parsedParams.data;
+        const { serverAdmin } = parsedBody.data;
+
+        const [existingUser] = await db
+            .select({
+                userId: users.userId,
+                serverAdmin: users.serverAdmin
+            })
+            .from(users)
+            .where(eq(users.userId, userId))
```
miloschwartz, line 97: make sure the user is internal user only

src/components/AdminUsersTable.tsx
```diff
@@ -369,11 +459,32 @@ export default function UsersTable({
                                         {t("generatePasswordResetCode")}
                                     </DropdownMenuItem>
                                 )}
+                                {!r.serverAdmin && (
```
miloschwartz, line 462: only show set/remove server admin option if user is internal (Not idp)

miloschwartz (approved):

PR #3750, opened 2026-09-15
Only accept http(s) targets for the resource auth redirect

kah-ja:
## AI Disclosure

The change was prepared with the help of Claude Code: the code reading, the patch and the checks below were done in that session and reviewed before opening this PR.

### What

`src/app/auth/resource/[resourceGuid]/page.tsx:121-150` copies the `redirect` query parameter into `redirectUrl` when `new URL(redirect).host` equals the resource host. The scheme is not looked at.

### Why

`URL` parses a host out of any scheme that uses `//`, so `javascript://resource-host/%0a...` has `host === "resource-host"` and passes the comparison. `redirectUrl` is handed to `ResourceAuthPortal` as the `redirect` prop and assigned to `window.location.href` after a successful login (`src/components/ResourceAuthPortal.tsx:213,247,281`), and `appendRequestToken` keeps the scheme when it appends the session parameter. The dashboard flows already go through the repo's own `cleanRedirect`/`sanitizeUrl` (`src/lib/cleanRedirect.ts:38`), which rejects `javascript:`, `data:` and protocol relative input; this page never called it because it needs an absolute URL on the resource host rather than an internal path.

### How

The target is parsed once into `redirectTarget` and the three branches that all assigned the same value are folded into one condition guarded by `isHttpTarget`. Every `http:`/`https:` target that was accepted before is still accepted, including the port and wildcard forms. `ResourceAuthPortal.appendRequestToken` and the identical helper in `src/components/AccessToken.tsx` are left unchanged; both receive their URL from the server component, which is where the value is now constrained.

## How to test?

- `npx tsc --noEmit` clean, `npx eslint` on the touched file clean, `npx prettier --check` clean (node 24, `npm ci`, `set:sqlite`, `set:oss`, `db:generate` as in `.github/workflows/test.yml`).
- Gate logic exercised directly with the resource URL `https://app.example.com/`: `javascript://app.example.com/%0a...` and `data://app.example.com/,x` were accepted before the change and are rejected after it; `https://app.example.com/dashboard`, `http://app.example.com:8080/x` stay accepted; `https://evil.example.net/` stays rejected.
- The repo has no test runner wired into CI (the `*.test.ts` files are standalone scripts), so no automated test was added for an app router page.

### References

GHSA-ffm9-x67r-gpw6 (reported privately, unpublished at time of writing)

Found during a [penetration test](https://www.turingpoint.de/en/security-assessments/pentests/) by turingpoint.


oschwartz10612 (approved):

oschwartz10612:
Thanks!

Issue #3761, opened 2026-09-16
Database backups: timestamps are month-1/zero-indexed, are not zero padded, and duplicate backups are generated


iconoclasthero:
### Describe the Bug

When I upgraded on Saturday, 2026-09-12, the backup file name is db_2026-8-12_20-35-56.sqite
```
# stat config/db/backups/db_2026-8-12_20-35-56.sqlite 
  File: config/db/backups/db_2026-8-12_20-35-56.sqlite
  Size: 21716992  	Blocks: 42416      IO Block: 4096   regular file
Device: 8,1	Inode: 529585      Links: 1
Access: (0644/-rw-r--r--)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2026-09-15 12:30:47.995058492 -0400
Modify: 2026-09-12 20:35:58.479893142 -0400
Change: 2026-09-12 20:35:58.479893142 -0400
 Birth: 2026-09-12 20:35:56.983870134 -0400
```
Furthermore, there are two non-identical backups, despite upgrading once:
```
.rw-r--r-- 22M 12 Sep 20:35 db_2026-8-12_20-35-56.sqlite
.rw-r--r-- 22M 12 Sep 20:36 db_2026-8-12_20-36-2.sqlite
```
```
# diff db_2026-8-12_20-3*
Binary files db_2026-8-12_20-35-56.sqlite and db_2026-8-12_20-36-2.sqlite differ
```
```
# stat db_2026-8-12_20-36-2.sqlite 
  File: db_2026-8-12_20-36-2.sqlite
  Size: 21716992  	Blocks: 42416      IO Block: 4096   regular file
Device: 8,1	Inode: 529753      Links: 1
Access: (0644/-rw-r--r--)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2026-09-15 12:30:47.719054280 -0400
Modify: 2026-09-12 20:36:03.308967432 -0400
Change: 2026-09-12 20:36:03.308967432 -0400
 Birth: 2026-09-12 20:36:02.985962465 -0400
```
Another user also confirmed this observation and that they have a
`db_2026-0-23_20-25-49.sqlite` created in January.

[...]

PR #3769, opened 2026-09-16
fix(setup): correct month index and zero-pad database backup file names

Hayyan612:

---

## Description

Fixes the month index and padding in database backup file names.

The timestamp was built inline from `Date#getMonth`, which is zero-indexed, so a backup taken on 12 September 2026 was written as `db_2026-8-12_20-35-56.sqlite`. No field was zero-padded either, which is where `db_2026-8-12_20-36-2.sqlite` came from.

The same expression existed in two places, and both are fixed:

- `server/setup/migrationsSqlite.ts:124`, in the `backupDb` helper
- `server/setup/scriptsSqlite/1.0.0-beta9.ts:37`, an inline copy of the same logic

Rather than correcting the expression twice, it moves to `formatBackupTimestamp` in `server/lib/backupFileName.ts` so the two paths cannot drift apart again. These were the only two `getMonth()` calls in the repo, and the Postgres setup path has no file backup, so nothing else is affected.

Padding every field has a useful side effect: the names now sort lexicographically into the order the backups were taken, which the old ones did not.

Before and after, using the names from the report:

| taken at | before | after |
|---|---|---|
| 12 Sep 2026 20:35:56 | `db_2026-8-12_20-35-56.sqlite` | `db_2026-09-12_20-35-56.sqlite` |
| 12 Sep 2026 20:36:02 | `db_2026-8-12_20-36-2.sqlite` | `db_2026-09-12_20-36-02.sqlite` |
| 23 Jan 2026 20:25:49 | `db_2026-0-23_20-25-49.sqlite` | `db_2026-01-23_20-25-49.sqlite` |

Existing backup files are not renamed. Only newly written ones get the corrected name.

## Issues closed by this PR

Partially addresses #3761, the naming half of it. See below for the duplicate-backup half.

## Additional Information

**Verification.** `npx tsc --noEmit` passes, which is what the test workflow runs. `server/lib/backupFileName.test.ts` follows the convention of the other tests in `server/lib` and runs with `npx tsx server/lib/backupFileName.test.ts`.

I checked the tests actually catch the bug rather than just passing alongside the fix. Swapping the helper body back to the old inline formula fails them with the exact name from the report:

```
Expected: 2026-09-12_20-35-56
Actual: 2026-8-12_20-35-56
```

**On the duplicate backups, which this PR does not change.** `backupDb()` is called inside the per-migration loop at `migrationsSqlite.ts:197`, so an upgrade takes one backup per pending migration rather than one per upgrade. Two pending migrations gives the two files about six seconds apart that @iconoclasthero saw. The `1.0.0-beta9` script also takes its own backup inline, on top of the loop's, whenever that migration runs.

That looks deliberate, a snapshot before each migration step, so I have left it alone rather than guess. Happy to follow up if you would rather it took one per upgrade.

One related risk worth flagging separately: the timestamp is second-granular and `fs.copyFileSync` overwrites silently, so two migrations completing inside the same second write to the same path and one backup is lost. Padding does not change that. Adding the migration version to the name would, if you want it.


oschwartz10612 (approved):

iconoclasthero:
Thanks for addressing this so quickly.  I was surprised last night when there were no other issues on the topic which I think is actually kudos to the Pangolin team that no one has apparently needed these backups!

> One related risk worth flagging separately: the timestamp is second-granular and fs.copyFileSync overwrites silently, so two migrations completing inside the same second write to the same path and one backup is lost. Padding does not change that. Adding the migration version to the name would, if you want it.

And it overwrites?  There's no chance the writes "collide" and corrupt the backup?

PR #3770, opened 2026-09-16
fix(ui): point the manual systemd unit at the installed CLI path

### LINKED ISSUE #3768: Instructions for manually create systemd service for site are faulty on 1.23.0
### Describe the Bug

Systemd service file creating instructions are faulty pointing not to the installed pangolin binary:

```
sudo tee /etc/systemd/system/pangolin-site.service > /dev/null << 'EOF'
[Unit]
Description=Pangolin Site
Wants=network-online.target
After=network-online.target

[Service]
Type=simple
User=root
Group=root
EnvironmentFile=/etc/pangolin/pangolin-site.env
ExecStart=/home/owen/fossorial/cli/bin/pangolin up site
Restart=always
RestartSec=2
UMask=0077

PrivateTmp=true

[Install]
WantedBy=multi-user.target
EOF
```

Should be /usr/local/bin/pangolin instead of /home/owen/fossorial/cli/bin/pangolin

### Environment

- Pangolin Version: 1.23.0
- Edition (Community or Enterprise): CE
- Gerbil Version: 1.5.1
- Traefik Version: 3.7
- Newt Version: N/a, pangolin CLI
- Client Version: 1.7.0

### To Reproduce

Check site creation instructions with manually creating the systemd service.

### AI Disclosure

None

### Expected Behavior

Systemd service file should be

```
sudo tee /etc/systemd/system/pangolin-site.service > /dev/null << 'EOF'
[Unit]
Description=Pangolin Site
Wants=network-online.target
After=network-online.target

[Service]
Type=simple
User=root
Group=root
EnvironmentFile=/etc/pangolin/pangolin-site.env
ExecStart=/usr/local/bin/pangolin up site
Restart=always
RestartSec=2
UMask=0077

PrivateTmp=true

[Install]
WantedBy=multi-user.target
EOF
```

Hayyan612:

---

## Description

The "Manual Systemd Service" instructions on the site install screen hand the user a unit file with a developer machine path baked into it:

```
ExecStart=/home/owen/fossorial/cli/bin/pangolin up site
```

Copy that as printed and the service fails to start, since nothing is at that path on a normal install.

`src/components/newt-install-commands.tsx:153` is the only occurrence in the repo.

I checked the right value against the installer the step immediately above it recommends rather than guessing. `https://static.pangolin.net/get-cli.sh`:

```sh
# Prefer /usr/local/bin for system-wide installation
echo "/usr/local/bin"
```

That also matches the rest of the block, where the other commands call bare `pangolin` and rely on it being on `PATH`.

If the CLI is installed somewhere else, for example via `get-cli.sh --path`, the unit still needs editing, but `/usr/local/bin/pangolin` is the correct default for the flow this screen describes.

## Issues closed by this PR

Closes #3768

## Additional Information

`npx tsc --noEmit` passes.


AstralDestiny:
Mmm good catch it should align with the docs. heh

oschwartz10612 (approved):

oschwartz10612:
Thanks!

PR #3772, opened 2026-09-16
fix(sites): fall back to newtVersion so the Type badge is never empty

### LINKED ISSUE #3766: Site overview: "Type" is empty when not on latest Newt version
### Describe the Bug

After upgrading to 1.23.0 EE (personal use), my site overview shows empty badges where it should show an outdated Newt version. If I update a site's Newt, the version shows correctly. See screenshot.

Image

### Environment

- Pangolin Version: 1.23.0
- Edition (Community or Enterprise): EE

### To Reproduce

1. Upgrade to 1.23.0
2. Open Site overview.

### AI Disclosure

No AI 

### Expected Behavior

Show all Newt versions, not only newest

Hayyan612:

---

## Description

Sites running a newt that has not reported the newer `agent` / `agentVersion` fields render an empty badge in the Type column.

`src/components/SitesTable.tsx` only bails out early when **both** fields are missing:

```tsx
if (!originalRow.agent && !originalRow.newtVersion) {
    return <span>-</span>;   // has not checked in yet
}
```

A site with `newtVersion` set but `agent` null therefore falls through to the badge, where the label was built solely from `agent` and the version solely from `agentVersion`:

```tsx
{originalRow.agent == "newt" ? "Newt" : null}
{originalRow.agent == "cli" ? "Pangolin CLI" : null}
{originalRow.agentVersion && <span>v{originalRow.agentVersion}</span>}
```

Both ternaries yield `null` and `agentVersion` is null, so the badge renders with nothing in it. That matches the screenshot on the issue, and it explains why updating a site's newt makes the version appear: the newer newt sends `agent`, which populates both columns.

This labels such a site `Newt` and falls back to `newtVersion` for the version, so the badge always carries the information it is there to show. The `updateAvailable` popup beside it already read `newtVersion`, so it was working while the badge next to it was blank.

## Issues closed by this PR

Closes #3766

## Additional Information

`npx tsc --noEmit` passes.

**One assumption worth your review.** I label a site with a null `agent` as `Newt` rather than leaving it blank. The reasoning is that `agent` is nullable and commented "either newt or cli" (`schema.ts:706`) while `version` predates it, so a row with a version but no agent is one that last checked in before the column was populated. If a Pangolin CLI site can also end up with a null `agent`, that assumption is wrong and it should render blank or fall back differently. Say the word and I will change it.

There is also a related upstream case I have deliberately left alone. `server/routers/newt/handleNewtRegisterMessage.ts:184-188` already tries to backfill:

```ts
agentVersion: !agentVersion && agent == "newt" ? newtVersion : agentVersion
```

The fallback is gated on `agent == "newt"`, so a newt that sends no `agent` at all never triggers it and both columns stay null. Defaulting `agent` there would repair the stored data rather than just the display, but it risks mislabelling CLI sites, so it seemed like your call rather than mine. Fixing the display also has the advantage of correcting existing rows immediately instead of waiting for the next check-in.


oschwartz10612 (approved):

oschwartz10612:
Thanks!

PR #3784, opened 2026-09-17
feat(resources): add multi-select bulk actions and date added column

orochibraru:

## AI Disclosure

Claude was used with the ponytail plugin. It handled the frontend integration as well as the translations. I also used it via Chrome MCP to run tests against a dev instance to make sure there's no flickering, misalignments or unexpected behaviour (in addition to manual testing of course). 

### Summary

Adds multi-select with bulk actions to the public resources table, plus a new "Date Added" column. I built this because while I don't use the Pangolin UI much I do automate pangolin entries via API through other tools. A pain point I'm often facing is trying to filter by the last resource created or simply deleting multiple resources or disabling auth on multiple resources at once. 

### Bulk actions
- Checkbox on each row and a select-all checkbox in the header (current page only)
- When rows are selected, an "N selected" menu appears next to the search box:
  - **Enable SSO Authentication**: sets `sso: true` on each selected resource. TCP/UDP resources are skipped since they have no auth; the action is disabled if only those are selected.
  - **Disable SSO Authentication**
  - **Delete Selected Resources**: uses the existing type-to-confirm delete dialog
- Requests are sent per resource with `Promise.allSettled`. Partial failures show a toast with the failed count, then the selection clears and the table refreshes.

### Date Added
- New nullable `createdAt` column on `resources` (sqlite and pg), set with `$defaultFn` on insert so every creation path (API, blueprints) is covered
- `1.24.0` migrations for sqlite (idempotent check) and pg (`ADD COLUMN IF NOT EXISTS`)
- `GET /org/:orgId/resources` returns `createdAt` and accepts `sort_by=createdAt`
- Column is sortable from its header, same as Name

## Notes
- Resources created before this change have no creation date to backfill, so they show `-`.
- The migration is versioned `1.24.0` and runs once the app version is bumped.
- New strings are added to `en-US.json` only.
- Private resources table is unchanged (no auth to enable there).

## Screenshots
01-date-added-column
02-bulk-actions-menu
02-rows-selected
03-bulk-actions-menu
04-bulk-delete-confirm
05-after-enable-sso
06-disable-sso-menu

## How to test?

Spin this up and head over to the public resources page. There you'll see the checkboxes next to each item to select them and trigger the select field appearance from which you'll be able to either delete multiple resources, enable or disable platform sso as well as the new "created at column".


AstralDestiny:
Curious what those 2 circles are from on those images.

orochibraru:
@AstralDestiny 
> Curious what those 2 circles are from on those images.

You mean the NextJS devtools on the bottom left and the Tanstack devtools on the bottom right?

Issue #3631, opened 2026-08-23
OIDC/Google/Microsoft identity provider buttons on login page show generic OIDC icon instead of branded logo


Bolex80:
### Describe the Bug

On the Pangolin login page, the SSO buttons for **Google** and **Microsoft (Azure)** identity providers both display the **generic OAuth2/OIDC icon** (`/idp/openid.png`) instead of their branded logos (`/idp/google.png` and `/idp/azure.png`). Both buttons previously showed the correct branded logo.

The brand information *is* present and correct in the data and in the `/idp` list API response — the login page simply ignores it when rendering the button list.

### Environment

- OS Type & Version: Ubuntu (Oracle Cloud A1, bolex-cloud)
- Pangolin Version: `fosrl/pangolin:ee-latest` (image built 2026-07-30)
- Edition (Community or Enterprise): **Enterprise**
- Gerbil Version: `fosrl/gerbil:latest`
- Traefik Version: `traefik:latest`
- Newt Version: N/A (server-side)
- Client Version: N/A

### To Reproduce

1. Create two OIDC identity providers, one configured with `variant: "google"` and one with `variant: "azure"` (the built-in Google and Azure templates).
2. Confirm in the `idpOidcConfig` table that `variant` is correctly set to `google` / `azure`, while `idp.type` is `oidc` (this is the intended model — `type` stays `oidc`, the brand lives in `variant`).
3. Open the login page (`/auth/login`).
4. Observe that both the Google and Microsoft buttons render the **same** generic OIDC icon.

### Expected Behavior

The Google button should show the Google logo (`/idp/google.png`) and the Microsoft button should show the Azure logo (`/idp/azure.png`), as they did previously.

### Actual Behavior

Both buttons render the generic OIDC icon (`/idp/openid.png`). The screenshot below shows both the Google and Microsoft buttons (note the "Last Used" badge on Google) displaying the generic key/identity icon rather than the branded logos.

![Pangolin login page showing generic OIDC icon on Google and Microsoft buttons](https://raw.githubusercontent.com/Bolex80/issue-screenshots/main/pangolin-login-idp-icons.png)

### Root Cause / Suggested Fix

The `/idp` list endpoint (`server/routers/idp/listIdps.ts`) already returns both `type` and `variant` correctly:

```ts
idpId: idp.idpId,
name: idp.name,
type: idp.type,          // "oidc" for all OIDC-backed IdPs
[...]

PR #3790, opened 2026-09-19
fix: use idp variant for resource auth login page icons

Bolex80:
## Bug Description

The **resource auth login page** (`auth/resource/[resourceGuid]`) — used for per-site / custom-branded logins like "Authenticate to access [resource]" — shows the **generic OIDC icon** on the Google and Microsoft SSO buttons instead of their branded logos.

The main login page and the org login page were already fixed, but the resource auth page's `else` branch was missed.

Fixes #3631

## Root Cause

`src/app/auth/resource/[resourceGuid]/page.tsx`, in the non-SaaS / non-org path (lines ~282-287):

```ts
const idpsRes = await priv.get<AxiosResponse<ListIdpsResponse>>("/idp");
loginIdps = idpsRes.data.data.idps.map((idp) => ({
    idpId: idp.idpId,
    name: idp.name,
    variant: idp.type   // <-- always "oidc" for OIDC-backed IdPs (Google, Azure)
})) as LoginFormIDP[];
```

This path passes `idp.type` as the icon `variant`. For OIDC-backed providers the brand lives in the separate `idpOidcConfig.variant` column (`"google"` / `"azure"`), while `idp.type` is always `"oidc"`. So `IdpTypeIcon` always resolves to the generic `/idp/openid.png`.

The `/idp` list API already returns `variant` (`server/routers/idp/listIdps.ts` selects both `type` and `variant`). This path simply wasn't using it.

## Fix

```ts
variant: idp.variant ?? idp.type
```

This mirrors the identical fix already applied to the main login page (`src/app/auth/login/page.tsx`) and the org login page (`src/app/auth/org/[orgId]/page.tsx`) in PR #3637 — the resource-auth page's `else` branch was the one place left unfixed. `idp.variant` is nullable in the API schema, so `?? idp.type` preserves the previous fallback behavior for providers without a branded variant.

## How to Verify

1. Configure Google and Microsoft/Azure OIDC identity providers (both stored with `idp.type = "oidc"` but `idpOidcConfig.variant = "google"` / `"azure"`).
2. Open a Pangolin-protected resource's login page (e.g. `https://<resource>/auth/login`, the branded "Authenticate to access X" page).
3. Confirm the Google button shows the multicolor Google logo and the Microsoft button shows the Azure logo instead of the generic OIDC icon.

## Test Plan

- [x] Added regression test for this bug — N/A (existing pattern; no unit test for this mapping on `main`; the fix matches the already-merged approach in PR #3637)
- [x] Existing tests still pass — the change is a one-line mapping, consistent with the shipped fix for `auth/login`
- [x] Manual verification — reproduced and confirmed the identical bug on the resource auth page

## Risk Assessment

Low — cosmetic-only change to which logo PNG is rendered. It does not affect authentication, redirect URLs, or IdP configuration. The fallback `?? idp.type` preserves current behavior for any provider where `variant` is null.


oschwartz10612:
Thanks!

oschwartz10612 (approved):

Issue #3720, opened 2026-09-10
AI Gateway: HTTP 302 from POST requests via OpenAI Python client (intercepted before reaching gateway)


KJELD-LRNT:
## Description

When AI clients using the OpenAI Python client (httpx-based) call the Pangolin AI Gateway endpoint, they intermittently receive HTTP 302 redirects. The OpenAI Python client does **not** follow 302 redirects for POST requests by default, so these 302s surface as API failures.

Critically, the 302 does **not** appear in Pangolin's `aiSessionLog` — across 2,667 logged requests, zero had a 302 status code. This strongly suggests the redirect is being intercepted **before** the request reaches the Pangolin AI Gateway, likely at the Cloudflare edge or Traefik middleware layer.

## Affected Clients

- **OpenAI Python client** (httpx) — does not follow 302 for POST by default
- **DBeaver AI Assistant** — identical issue reported at [dbeaver/dbeaver#41932](https://github.com/dbeaver/dbeaver/issues/41932) (Aug 26, 2026)
- Likely any HTTP client that does not automatically follow redirects on POST requests

## Environment

- Pangolin version: 1.22+
- AI Gateway resource configured with a custom OpenAI-compatible provider
- HTTPS endpoint (port 443), Cloudflare DNS configured in DNS-only mode (not proxied through Cloudflare CDN)
- Traefik reverse proxy in front of Pangolin

## Steps to Reproduce

1. Configure an AI Gateway resource in Pangolin with an OpenAI-compatible provider
2. Point the OpenAI Python client (`openai` package) at the gateway endpoint
3. Make API calls intermittently (especially at the start of a worker run)
4. Observe that all 3 retries fail within ~10 seconds, each receiving a 302

## Investigation Findings

| Evidence | Detail |
|----------|--------|
| `aiSessionLog` | Zero 302 status codes across 2,667 logged requests |
| Docker / Traefik logs | No 302 entries matching the failed requests |
| Direct `curl` | Always returns HTTP 200 — works fine |
| Transient | 302 resolves on its own within minutes |
| Client behavior | All 3 retries fail with 302 in <10 seconds |

## Expected Behavior

POST requests to the AI Gateway should be forwarded to the gateway without being intercepted by intermediate layers (Cloudflare edge, Traefik middleware). When a 302 is returned, it should be logged in the `aiSessionLog` if it originated from Pangolin.

## Actual Behavior

[...]

PR #3797, opened 2026-09-21
fix(ai-gateway): return JSON auth error instead of 302 for API clients on inference resources

MauryaQbit:

Fixes #3720.

Problem
POSTs to AI Gateway inference resources via the OpenAI Python client (httpx) intermittently fail with HTTP 302. Nothing appears in aiSessionLog because Badger issues the redirect before the request reaches the gateway. httpx does not follow 302 on POST, so the SDK surfaces it as a hard failure. Same symptom was reported for DBeaver in dbeaver/dbeaver#41932.

Cause
verify-session classifies browser vs API client with a UA blocklist. OpenAI/Python and python-httpx were not in the list, so those SDKs were treated as browsers and given a redirectUrl to /auth/resource/..., which Badger turns into 302 Found. curl worked because curl/* is blocklisted and gets a JSON 401 instead.

Change
- Extend the non-browser UA list to cover httpx/httpcore/openai/anthropic/dbeaver and other JVM/Python transports.
- Never redirect programmatic API calls on inference resources: non-GET, known capability paths, or non-HTML Accept now always get the capability-shaped 401 JSON.
- Case-insensitive header lookup for User-Agent/Accept/Sec-Fetch-Mode.

Browser navigation (GET + text/html) is unchanged and still redirects to the resource auth page.

Testing
- UA matrix checked: OpenAI/Python, python-httpx, DBeaver, Java, curl -> non-browser; Chrome -> browser.
- POST /v1/chat/completions with bad/missing VAK now returns 401 JSON instead of 302 regardless of UA.
- GET /v1/models with bad VAK returns 401 JSON; GET / in a browser still redirects.


MauryaQbit:
Hey @miloschwartz  can you review this PR

oschwartz10612 (approved):

oschwartz10612:
Thanks

